PR #1, opened 2023-09-01
build: add nix flake for linting + devShell + workflow

mrcjkb:
I noticed there's a stylua run in the makefile, but there's no `.stylua.toml`.

- [x] That should probably be added before this gets merged.

This adds a Nix flake with:

- A devShell with things like `lua-language-server`, `luacheck`, `stylua`, ...
- `checks` that run the linters (I also included `editorconfig-checker`, but I'm not sure if we want that?).
- A GitHub workflow that runs the checks on push and PR.

If we decide to merge this, I would like to follow-up with a `check` that runs `busted` tests using [`neorocksTest`](https://github.com/nvim-neorocks/neorocks#in-a-nix-flake).


mrcjkb:
Note: When run locally, the `stylua` check fails.

mrcjkb:
Oh wait - there is a `stylua.toml` (not hidden) :thinking: I wonder why it's failing...

vhyrro:
It's likely because you're not compiling stylua with 5.2 support, which is required for parsing of `goto` and some other stuff. With cargo it's as simple as `cargo install stylua --features lua52` but unsure how it works in nix.

vhyrro:
Also a stylua.toml is also considered standard and it picks it up on my end with no problems :thinking:

mrcjkb:
Let's see if the fix gets merged on the Nix side: https://github.com/cachix/pre-commit-hooks.nix/pull/346
The stylua behaviour is intended.

nix/plugin-overlay.nix
```diff
@@ -0,0 +1,9 @@
+{
+  name,
+  self,
+}: final: prev: {
+  vimPlugins.rocks-nvim = final.pkgs.vimUtils.buildVimPluginFrom2Nix {
```
teto, line 5: careful, in latest nixpkgs, it's now `buildVimPlugin` ;)

teto (approved):

nix/plugin-overlay.nix
mrcjkb: Hmm, maybe we should use `buildLuarocksPackage` instead.
It's gonna need toml-edit and some other dependencies.

PR #2, opened 2023-10-08
refactor: type annotations + formatting

mrcjkb:
Just some type annotation + formatting refactors I did while getting to know the code base :smile: 

Doesn't change any behaviour.


lua/rocks/internal-types.lua
```diff
@@ -0,0 +1,22 @@
...
+--
+---@brief [[
+--
+-- rocks.nvim internal type definitions.
+--
+---@brief ]]
+
+---@class (exact) Rock
+---@field public name string
+---@field public version string
```
mrcjkb, line 20: This is a WIP feature of lua-language-server that will allow more strict type definitions (e.g., creating records with undefined fields would be invalid).

It doesn't work properly yet, but it will likely turn out to be useful in the future.

lua/rocks/init.lua
```diff
@@ -9,21 +9,18 @@
...
+---@brief ]]
 
 local rocks = {}
 
 local setup = require("rocks.setup")
 local config = require("rocks.config")
 
+---@param opts RocksOptions
 function rocks.setup(opts)
     assert(vim.version() >= vim.version.parse("0.10.0-dev"), "rocks.nvim requires Neovim 0.10.0 or later!")
```
mrcjkb, line 25: Do we want to get rid of `setup`?

I would suggest to keep the `config` module and add the following:



```lua
--- @type RocksOptions
vim.g.rocks_nvim = vim.g.rocks_nvim -- for lua-language-server

---@type RocksConfig
local default_config = {
    ---@diagnostic disable-next-line: param-type-mismatch
    rocks_path = vim.fs.joinpath(vim.fn.stdpath("data"), "rocks"),
    ---@diagnostic disable-next-line: param-type-mismatch
    config_path = vim.fs.joinpath(vim.fn.stdpath("config"), "rocks.toml"),
}

local opts = validate(vim.g.rocks_nvim or {}) -- validate not implemented yet
---@type RocksConfig
local config = vim.tbl_deep_extend("force", config, opts or {})
return config
```

And to move the code in the `setup` module to the `plugin` directory.

vhyrro:
Awesome work :D

PR #4, opened 2023-10-08
fix(sync): do not attempt to remove dependencies

### LINKED ISSUE #3: sync: do not attempt to remove dependencies
## Summary

If a plugin that other plugins depend on is removed from rocks.toml,
rocks.nvim attempts to remove it.

## Spec

Given the following rocks.toml:

```toml
[plugins]
"haskell-tools.nvim" = "2.3.0" # depends on plenary.nvim
"plenary.nvim" = "0.1.3"
```
And both plugins have been installed with `:Rocks install`
When I remove plenary.nvim from the rocks.toml as follows:
```toml
[plugins]
"haskell-tools.nvim" = "2.3.0" # depends on plenary.nvim
```
And I execute `:Rocks sync`

### Expected behaviour
Then nothing happens

### Current behaviour
Then the rocks.nvim displays "Removing 'plenary.nvim'"

## Implementation details

- `luarocks` does not have a way to list dependencies, but `luarocks show <rock_name>` will list the dependencies of each rock.
- We could use this on each rock in `rocks.toml` to gather all dependencies, and then filter the list of rocks to remove accordingly.

mrcjkb:
Note: This is stacked on top of #2.
Because #2 is from my fork, I can't point this PR to that branch.

Fixes #3.

- [ ] Needs to be tested

### DIFF
--- a/lua/rocks/internal-types.lua
+++ b/lua/rocks/internal-types.lua
@@ -19,4 +19,8 @@
 ---@field public name string
 ---@field public version string
 
+---@class (exact) RockDependency
+---@field public name string
+---@field public version? string
+
 --- internal-types.lua ends here
--- a/lua/rocks/luarocks.lua
+++ b/lua/rocks/luarocks.lua
@@ -1,0 +1,40 @@
+---@mod rocks.luarocks
+--
+-- Copyright (C) 2023 NTBBloodbath
+--
+-- Version:    0.1.0
+-- License:    GPLv3
+-- Created:    19 Jul 2023
+-- Updated:    27 Aug 2023
+-- Homepage:   https://github.com/nvim-neorocks/rocks.nvim
+-- Maintainers: NTBBloodbath <bloodbathalchemist@protonmail.com>, Vhyrro <vhyrro@gmail.com>
+--
+---@brief [[
+--
+-- Functions for interacting with the luarocks CLI
+--
+---@brief ]]
+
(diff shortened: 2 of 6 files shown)


vhyrro:
Pull `main` in your fork and rebase on that, gotta get rid of those conflicts 😅

mrcjkb:
@vhyrro I haven't tested this at all yet (not even manually) - I still have to add a Neovim derivation that uses `rocks.nvim` to the nix flake, so I can test these kinds of things.

But I would also like to add an intergration test for it.

NTBBloodbath (approved):
It looks good to me, I will test the changes using `image.nvim` rock. If everything goes well, can we merge or do you want to wait until you do the derivation with nix for testing first?

Off-topic: I think we should change the copyright notice in the comments header, maybe use the organization's name instead of mine? _This would be another PR, of course_

mrcjkb:
> It looks good to me, I will test the changes using `image.nvim` rock. If everything goes well, can we merge or do you want to wait until you do the derivation with nix for testing first?

Thanks :smile: 
I finished it quite late yesterday. If it works, I can add the test in a follow-up PR.

> Off-topic: I think we should change the copyright notice in the comments header, maybe use the organization's name instead of mine? _This would be another PR, of course_

Sure, that's fine by me.

vhyrro:
All fixes are in and it all looks good! :D

PR #6, opened 2023-10-23
refactor(commands): use table to map commands to implementations

mrcjkb:
This lets us reuse the keys for the completions.
In a follow-up, we could change the `command_tbl` to something like this:

```lua
command_tbl = {
  install = {
     command = function(_) ...end,
     completions = function(arg_lead) 
       -- run `luarocks query`
     end,
   },
}
```

(if it can be done without lagging neovim)


vhyrro:
LGTM!

vhyrro (approved):

PR #7, opened 2023-10-23
tests: add integration test

mrcjkb:
For now, I have set `--option sandbox false`, to work around https://github.com/luarocks/luarocks/issues/1539 (which is not ideal).


.editorconfig
```diff
@@ -0,0 +1,53 @@
...
+trim_trailing_whitespace = true
+
+[*.nix]
+indent_size = 2
+
+# Ignore paths
+[lua/nio/**]
+charset = unset
+end_of_line = unset
+insert_final_newline = unset
```
teto, line 17: I am not familiar with the `unset`. What's the advantage of defining those ?

Makefile
```diff
@@ -1,5 +1,5 @@
 format:
-	stylua -v --verify lua/rocks/ plugin/
+    stylua -v --verify lua/rocks/ plugin/
```
teto, line 2: seems like a mistake

.editorconfig
mrcjkb: editorconfig doesn't have an "ignore" config.
This is a workaround for that.

Makefile
mrcjkb: I think that was editorconfig complaining

teto: I mean dont makefiles only accept tab before commands ?

teto (approved):

Makefile
mrcjkb: You're probably right ..

.luacheckrc
```diff
@@ -1,6 +1,10 @@
 globals = {
     "_",
-    "vim"
+    "vim",
+    "describe",
+    "it",
+    "assert",
+    "stup",
```
NTBBloodbath, line 7: I'm not sure why this was added to the globals, maybe it was a typo? I couldn't find anything with that name in the code

NTBBloodbath (approved):
LGTM

PR #9, opened 2023-10-24
build: neovim derivation with rocks.nvim for manual testing with nix

mrcjkb:
Currently, you have to call `:lua require('rocks').setup()` for the plugin to initialize.

### FIXMEs:

- [x] We need to get rid of the `setup` function (separate PR)
- [x] I had to add `toml` and `toml-edit` to the `LUA_CPATH` via `wrapperArgs`(@teto I thought `buildNeovimPlugin` adds the dependencies?).
- [ ] Haven't been able to test `:Rocks install` yet.

### DIFF
--- a/nix/plugin-overlay.nix
+++ b/nix/plugin-overlay.nix
@@ -1,9 +1,68 @@
 {
   name,
   self,
-}: final: prev: {
-  vimPlugins.rocks-nvim = final.pkgs.vimUtils.buildVimPlugin {
-    inherit name;
+}: final: prev: let
+  lib = final.lib;
+  rocks-nvim-luaPackage-override = luaself: luaprev: {
+    rocks-nvim = luaself.callPackage ({
+      luaOlder,
+      buildLuarocksPackage,
+      lua,
+      toml,
+      toml-edit,
+    }:
+      buildLuarocksPackage {
+        pname = name;
+        version = "scm-1";
+        knownRockspec = "${self}/rocks.nvim-scm-1.rockspec";
+        src = self;
+        disabled = luaOlder "5.1";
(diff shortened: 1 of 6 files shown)


teto:
>  I had to add toml and toml-edit to the LUA_CPATH via wrapperArgs(@teto I thought buildNeovimPlugin adds the dependencies?).

I think that's a bug in nixpkgs, the hook may not work correctly

nix/plugin-overlay.nix
```diff
@@ -1,9 +1,68 @@
...
+      plugins = with final.vimPlugins; [
+        rocks-nvim
+      ];
+    };
+    runtimeDeps = with final; [
+      lua5_1
+      luarocks
+    ];
+  in
+    final.wrapNeovimUnstable final.neovim-nightly (neovimConfig
```
teto, line 50: pretty cool. Hopefully ithe interface will become easier after 23.11. I prefer to keep the wrapperArgs as a list and was thinking of making it mandatory in next nixpkgs (but I can't justify it yet so maybe wont).

nix/plugin-overlay.nix
```diff
@@ -1,9 +1,68 @@
...
+      // {
+        wrapperArgs =
+          lib.escapeShellArgs neovimConfig.wrapperArgs
+          + " "
+          + ''--set NVIM_APPNAME "nvimrocks"''
```
teto, line 55: pretty cool usage of NVIM_APPNAME

teto (approved):

nix/plugin-overlay.nix
mrcjkb: It's a really nice feature, especially for Nix...

PR #10, opened 2023-10-24
feat!: auto-setup + healthchecks + generate vimdocs

mrcjkb:
- Replace `setup` with smart initialization
- Validate config
- Use `lemmy-help` to generate vimdoc from luadoc
- Add health checks

- [ ] We may have to add 
  - [ ] a `CACHIX_AUTH_TOKEN`
  - [ ] and a `GITHUB_TOKEN`
  for the `docgen` workflow.


lua/rocks/health.lua
```diff
@@ -0,0 +1,183 @@
...
+local external_dependencies = {
+    {
+        name = "luarocks",
+        get_binaries = function()
+            return { "luarocks" }
+        end,
+        optional = function()
+            return true
+        end,
+        url = "[laurocks](https://luarocks.org/#quick-start)",
```
NTBBloodbath, line 74: There is a small typo :)

NTBBloodbath (approved):
LGTM!

Just a small question, `vim.g.rocks_nvim` would return a table when used as a function, right?

PR #11, opened 2023-10-25
feat: name and version completions for `:Rocks install`

mrcjkb:
- package names
rocks-query-1

- package versions
rocks-query-2


NTBBloodbath (approved):
LGTM, the truth is I never thought that something like this would be possible (especially having a very good performance), awesome work bud! :)

vhyrro:
Time to try this out and see if it works :crossed_fingers:

vhyrro:
Holy moly it's fast. Works on my end!

PR #15, opened 2023-10-28
Readme overhaul

NTBBloodbath:
Now that the MVP is closer to being a reality, it's time to start refactoring the README to reflect the current state of the project _and make it more fancy, of course_.

This is a draft and should be merged **only after** the installer is ready and the documentation information is ready. Any feedback or suggestion is welcome!


README.md
```diff
@@ -1,4 +1,8 @@
-# `rocks.nvim` (WIP)
+[<img src="https://raw.githubusercontent.com/luarocks/luarocks-logos/master/luarocks_logo_only.png" align="right" width="144" />](https://luarocks.org/)
+<img src="https://pngfre.com/wp-content/uploads/heart-87-1024x849.png" align="right" width="104" />
+[<img src="https://neovim.io/logos/neovim-mark-flat.png" align="right" width="114" />](https://neovim.io/)
+
+# `rocks.nvim`
```
mrcjkb, line 5: suggestion: Personally, I like to use HTML for readme headings (see for example the [luarocks-tag-release README](https://github.com/nvim-neorocks/luarocks-tag-release), and to spell out the plugin name as an SVG image (Lato font - same as Neovim uses), with "rocks" slim and ".nvim" bold.

(but that's just personal taste). If you like, I can play around with the image in Inkscape.

README.md
```diff
@@ -9,20 +13,44 @@ A modern approach to Neovim plugin management.
 
 ## :star2: Features
 
-- `Cargo`-like `rocks.toml` file for declaring all your plugins
-- Name-based installation (`use "nvim-neorg/neorg"` becomes `:Rocks install neorg` instead)
-- Automatic dependency and build script management
+- `Cargo`-like `rocks.toml` file for declaring all your plugins.
+- Name-based installation (`use "nvim-neorg/neorg"` becomes `:Rocks install neorg` instead).
```
mrcjkb, line 17: suggestion: since packer is practically dead at this point

```suggestion
- Name-based installation (` "nvim-neorg/neorg" ` becomes `:Rocks install neorg` instead).
```

README.md
```diff
@@ -9,20 +13,44 @@ A modern approach to Neovim plugin management.
...
 - True semver versioning!
-- (WIP) Automatic running of test suites
+- (WIP) Automatic running of test suites.
+
+## :pencil: Requirements
+
+- An up-to-date `Neovim` nightly (>= 0.10) installation.
+- A C/C++ compiler of your choice (must be compatible with `C++17`!) + [`CMake`](https://cmake.org/).
+- [`rust`](https://www.rust-lang.org/) toolchain (recommended latest stable version).
+- [`luarocks`](https://luarocks.org/) installed on your system and accessible in your shell.
```
mrcjkb, line 27: thought: maybe we should consider [hosting our dependencies as binary rocks](https://github.com/luarocks/luarocks/wiki/Hosting-binary-rocks).

README.md
```diff
@@ -9,20 +13,44 @@ A modern approach to Neovim plugin management.
...
 
-## Installation
+## :hammer: Installation
+
+### :snowflake: Nix
 
 <!-- Nix users have a skill issue here what else can I say.
 Just joking - add an entry for nix users as well. -->
 
+TODO...
```
mrcjkb, line 36: I don't think many Nix users will use a neovim plugin as a plugin manager.
And it's not worth polluting the readme for the few that will (we can use the wiki for that if we need to).

README.md
NTBBloodbath: Sounds good, I like the idea. I would definitely like to see it :)

README.md
NTBBloodbath: Done!

README.md
NTBBloodbath: I didn't know something like this was possible. We should definitely do this, it would reduce dependencies and make the process much faster, I'll take a look at it :p

PR #23, opened 2023-11-15
ci: add static type checking for neovim-nightly

mrcjkb:
This adds static type checking for neovim-nightly to our CI.

Note that I didn't add it as a pre-commit-hook to the Nix shell, because it can be quite slow.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -58,6 +58,59 @@
           ];
         };
 
+        mkTypeCheck = {
+          nvim-api ? [],
+          disabled-diagnostics ? [],
+        }:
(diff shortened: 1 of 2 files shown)


teto:
isn't this something that should be done in neovim ? what's the advantage of doing it in this repo ?

mrcjkb:
> isn't this something that should be done in neovim ? what's the advantage of doing it in this repo ?

We're type checking this repo against the neovim nightly API.

mrcjkb:
@teto are you okay if we merge this?

teto (approved):

PR #28, opened 2023-11-20
ci(luarocks): fix

mrcjkb:


mrcjkb:
It looks like [toml-edit has never been published using the luarocks-tag-release action](https://github.com/vhyrro/toml-edit.lua/actions/workflows/luarocks.yml).

I think it's something to do with missing toolchains or toolchains not being detected properly. But I'll have to pick this up later...

PR #30, opened 2023-11-24
[FEATURE]: Implement an install script for `rocks.nvim`

vhyrro:
This PR adds a fully working installation process using a remote `:source` command invocation.

While there are still a few rough edges I believe it's a good enough MVP to push into main :)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -61,3 +61,11 @@ insert_final_newline = unset
 trim_trailing_whitespace = unset
 indent_style = unset
 indent_size = unset
+
+[installer.lua]
+charset = unset
+end_of_line = unset
+insert_final_newline = unset
+trim_trailing_whitespace = unset
+indent_style = unset
+indent_size = unset
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.0.1](https://github.com/nvim-neorocks/rocks.nvim/compare/v1.0.0...v1.0.1) (2023-11-22)
+
+
+### Bug Fixes
+
+* **operations:** only find dependencies of rocks that exist ([#21](https://github.com/nvim-neorocks/rocks.nvim/issues/21)) ([6d3810d](https://github.com/nvim-neorocks/rocks.nvim/commit/6d3810dfaa3eabd9a23fa433e0f308fc7b16102d))
+
 ## 1.0.0 (2023-11-20)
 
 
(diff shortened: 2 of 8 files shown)


mrcjkb (commented):
question: Do we need to configure luarocks to use the rocks-binaries so that rocks.nvim can update itself?
Otherwise, users would have to have the rust and C++ toolchains installed, wouldn't they?

vhyrro:
Yep, without the binary rocks users would need to have a modern CPP compiler and rust tool chain. If luarocks can't find the binary rock for the target architecture it'll try compiling instead.

NTBBloodbath (approved):
LGTM!

NTBBloodbath:
> question: Do we need to configure luarocks to use the rocks-binaries so that rocks.nvim can update itself?
> Otherwise, users would have to have the rust and C++ toolchains installed, wouldn't they?

How could we do that? Since it is the dependencies that we want to extract from rocks-binaries and not the complete package 🤔

mrcjkb:
> > question: Do we need to configure luarocks to use the rocks-binaries so that rocks.nvim can update itself?
> > Otherwise, users would have to have the rust and C++ toolchains installed, wouldn't they?
> 
> How could we do that? Since it is the dependencies that we want to extract from rocks-binaries and not the complete package 🤔

Maybe we have to specify the dependencies separately - or we can add rocks.nvim to the rocks-binaries manifest, but not as a binary. And add a `server` option to the `rocks.toml` settings or something.

vhyrro:
Not sure if I understand the questions being asked here - the server that we configure is is simply the *preferred* server. When luarocks tries to pull in `rocks.nvim`, it notices the dependencies, then first tries to pull from our server before falling back to the regular luarocks manifest. Everything works without interruptions :)

mrcjkb:
> Not sure if I understand the questions being asked here - the server that we configure is is simply the _preferred_ server. When luarocks tries to pull in `rocks.nvim`, it notices the dependencies, then first tries to pull from our server before falling back to the regular luarocks manifest. Everything works without interruptions :)

That's the install script. What about updates? I don't think we specify `--server` for `Rocks sync`?

vhyrro:
Ah you're right. I was certain I also added that, but apparently I didn't. All rocks operations should also pass the `--server` flag

mrcjkb:
I just tried running the installer (from a local file).

- There was a graphical glitch when editing the installation path:

installer

- I was unable to change the `setup_luarocks` option from `true` to `false` (`E21: Cannot make changes, 'modifiable' is off`) and it had a similar graphical glitch.

vhyrro:
Hm, weird. I guess it has to do with some calculations that the installer is making. Let me see if I can fix that.

vhyrro:
Should be fixed now! :D

mrcjkb (approved):

PR #36, opened 2023-11-28
feat: allow users to install development rocks (`scm-1`)

NTBBloodbath:
The following ways are valid:

- `nvim-treesitter = "scm-1"` in your `rocks.toml`
- `:Rocks install nvim-treesitter dev`, operations will get `dev` version and convert it to `--dev` flag when installing the rock

### DIFF
--- a/lua/rocks/operations.lua
+++ b/lua/rocks/operations.lua
@@ -40,7 +40,12 @@ operations.install = function(name, version)
         name,
     }
     if version then
-        table.insert(install_cmd, version)
+        -- If specified version is dev then install the `scm-1` version of the rock
+        if version == "dev" then
+            table.insert(install_cmd, 2, "--dev")
+        else
+            table.insert(install_cmd, version)
+        end
     end
     local systemObj = luarocks.cli(install_cmd, function(obj)
         if obj.code ~= 0 then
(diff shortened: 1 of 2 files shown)


lua/rocks/operations.lua
```diff
@@ -145,7 +150,14 @@ operations.sync = function(user_rocks)
                 text:render_char(split.bufnr, -1, linenr_copy, 0, linenr_copy, msg_length)
 
                 table.insert(actions, function()
-                    local ret = operations.install(user_rocks[key].name, user_rocks[key].version).wait()
+                    -- If the plugin version is a development release then we pass `dev` as the version to the install function
+                    -- as it gets converted to the `--dev` flag on there, allowing luarocks to pull the `scm-1` rockspec manifest
+                    local ret
+                    if user_rocks[key].version == "scm-1" then
```
mrcjkb, line 156:
```suggestion
                    if vim.startswitn(user_rocks[key].version, "scm-") then
```

lua/rocks/operations.lua
```diff
@@ -309,6 +321,10 @@ operations.add = function(rock_name, version)
                 user_rocks.plugins = {}
             end
 
+            -- Set installed version as `scm-1` if development version has been installed
+            if version == "dev" then
+                installed_rock.version = "scm-1"
+            end
```
mrcjkb, line 327: question: What if the user called `:Rocks install <plugin> dev`? Does that result in the version being saved as `scm-1`?
If so, I think just saving it as `"dev"` is fine.

mrcjkb (approved):

lua/rocks/operations.lua
mrcjkb: Oops, typo on my end. Should have been "startswith"

lua/rocks/operations.lua
NTBBloodbath: > What if the user called `:Rocks install <plugin> dev`? Does that result in the version being saved as `scm-1`?

Yup, the plugin is stored as `<plugin> = "scm-1"` in the `rocks.toml` file. If the version is saved as `dev` instead it would cause errors when trying to update or sync since the regex we use to check the versions will not match anything

lua/rocks/operations.lua
NTBBloodbath: fixed :p

NTBBloodbath:
Can we merge this or do we need changes? The only way to install cmp and gitsigns is using the `scm-1` version as they don't have an automatic releases CI for luarocks xD

mrcjkb:
Go ahead and merge.

I typically use [conventional comments](https://conventionalcomments.org/) for code reviews and add a prefix (+ don't approve) if something is blocking.

PR #44, opened 2023-12-01
feat: add `:Rocks edit` command for opening `rocks.toml`

mrcjkb:
Note: This is stacked on #41, to avoid merge conflicts

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,11 @@ that are no longer needed, run the `:Rocks prune [rock]` command.
 > - The command provides completions for rocks that can safely
 >   be pruned without breaking dependencies.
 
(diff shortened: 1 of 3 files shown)


NTBBloodbath:
I'm not so sure about the name, maybe `config` or `edit` would make more sense for being more idiomatic? 🤔

mrcjkb:
> I'm not so sure about the name, maybe `config` or `edit` would make more sense for being more idiomatic? 🤔

How about `editToml`?

vhyrro:
I think just `edit` is the way to go. `:Rocks edit` -> " edit my rocks"

NTBBloodbath (approved):

Issue #48, opened 2023-12-03
`rocks.api` module that rocks can use to extend rocks.nvim


mrcjkb:
The idea would be a `rocks.api` module that exposes functions for

- [x] (fuzzy?-) searching installed rocks
- [x] (fuzzy?-) searching rocks on luarocks
- [x] Adding subcommands to `:Rocks`
- [x] Get `rocks.toml`

so plugins could integrate with rocks.nvim, e.g. with a telescope picker.

PR #53, opened 2023-12-07
perf: populate luarocks.org state cache at startup

mrcjkb:
#### Manual test steps:

```console
nix run .#neovim-with-rocks
```

```vim
:Rocks install <tab>
```

-> instant completions

The startup impact isn't really noticeable, but I've added a `vim.g.rocks_nvim.lazy` option that is set to `false` by default.

This also adds a small refactor:

- The rocks are now stored in a `{ [string]: Rock[] }` table (indexed by name), in preparation for #48.


NTBBloodbath (approved):

NTBBloodbath (approved):
For some reason GitHub showed only two files to me while reviewing the first time haha

LGTM!

Issue #55, opened 2023-12-09
[Feature]: Keep track of luarocks errors


mrcjkb:
Currently, operations like `Rocks sync` fail if a single `luarocks` CLI operation fails (e.g. when `Future.wait()` errors).

Instead, we should

- Keep track of the errors, continuing if one fails
- Log errors to a log file that can be opened with `:Rocks log`
- Report if any errors have occurred (e.g. with the message `Sync completed with errors. Run ':Rocks log'` for details`).
- Or, at the end of an operation, in the event of an error, ask the user if they would like to open the log file.

PR #58, opened 2023-12-09
feat(ui): use fidget.nvim for progress reports

mrcjkb:
This uses the new `fidget.progress` API for reporting operations progress.

It feels a bit more modern than the current `nui` implementation, with the ability to report progress as a percentage (`update`, `sync`, ...), and show signs of life with the spinner animation.

tty

For now, error handling isn't that advanced.
I plan on keeping track of errors with a log file (#55).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,7 @@ indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
 
-[*.{nix,md}]
+[*.{nix,md,json}]
 indent_size = 2
 
 [Makefile]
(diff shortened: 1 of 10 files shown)


mrcjkb:
Looks like luarocks.org is down right now (causing the fidget-nvim derivation to fail to build).

vhyrro:
No idea why the checks are failing, but it seems to simply be unable to pull data from the internet. Doesn't seem like anything huge. This PR, however, is huge! Code looks good and it seems to work great when I tested it. :purple_heart:

PR #70, opened 2023-12-11
perf: auto-populate removable rocks cache

mrcjkb:


mrcjkb:
Merging without review, as this is rather trivial.

PR #82, opened 2023-12-12
feat: auto add newly installed plugins to the RTP

vhyrro:
This PR adds freshly installed plugins to the runtimepath so they can be used right away.

TODOs:
- [x] Make this configurable within the rocks options

This approach theoretically adds more than usually required to the RTP (as by default we use a glob over all installed plugins), however this is only for the current session (which is a good tradeoff in my opinion) and makes sure everything works smoothly.

### DIFF
--- a/lua/rocks/operations.lua
+++ b/lua/rocks/operations.lua
@@ -97,6 +97,9 @@ operations.install = function(name, version, progress_handle)
             if progress_handle then
                 progress_handle:report({ message = message })
             end
+
+            vim.opt.runtimepath:append(vim.fs.joinpath(config.rocks_path, "lib", "luarocks", "rocks-5.1", "*", installed_rock.name:lower()))
+
             future.set(installed_rock)
         end
     end)
@@ -305,6 +308,8 @@ operations.sync = function(user_rocks)
                 percentage = get_progress_percentage(),
                 message = is_downgrading and ("Downgraded: %s"):format(key) or ("Upgraded: %s"):format(key),
             })
+
+            vim.opt.runtimepath:append(vim.fs.joinpath(config.rocks_path, "lib", "luarocks", "rocks-5.1", "*", user_rocks[key].name:lower()))
         end
 
         -- Determine dependencies of installed user rocks, so they can be excluded from rocks to prune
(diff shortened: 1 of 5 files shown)


lua/rocks/operations.lua
```diff
@@ -97,6 +97,9 @@ operations.install = function(name, version, progress_handle)
             if progress_handle then
                 progress_handle:report({ message = message })
             end
+
+            vim.opt.runtimepath:append(vim.fs.joinpath(config.rocks_path, "lib", "luarocks", "rocks-5.1", "*", installed_rock.name:lower()))
```
mrcjkb, line 101: issue: I think putting this in `operations.install` will add plugins to the `rtp` when we upgrade/downgrade a plugin, which I'm not sure we want.
Maybe it should go in `operations.add` instead.

lua/rocks/operations.lua
```diff
@@ -305,6 +308,8 @@ operations.sync = function(user_rocks)
                 percentage = get_progress_percentage(),
                 message = is_downgrading and ("Downgraded: %s"):format(key) or ("Upgraded: %s"):format(key),
             })
+
+            vim.opt.runtimepath:append(vim.fs.joinpath(config.rocks_path, "lib", "luarocks", "rocks-5.1", "*", user_rocks[key].name:lower()))
```
mrcjkb, line 312: issue: I think we only want to do this for newly added plugins.

lua/rocks/operations.lua
vhyrro: I actually specifically added it here because in theory a downgraded plugin could have a vastly different directory structure (or different rtp paths altogether) which won't be refreshed upon changing the plugin. Changing vim.opt forces a refresh internally :)

lua/rocks/operations.lua
vhyrro: I just realized this bit is  extraneous as it's done in the install function anyway

mrcjkb (approved):

mrcjkb (approved):

NTBBloodbath (approved):

PR #83, opened 2023-12-14
chore(nix): use test suite as base for devShell

mrcjkb:
This allows us to run `luarocks test` with Neovim as the interpreter from within the devShell.


mrcjkb:
- [ ] TODO: Disable checks in devShell

Issue #81, opened 2023-12-12
[Module]: `rocks-git` for pulling plugins that don't exist on luarocks


vhyrro:
Not everyone must have their plugins present on luarocks. This module would allow users to specify a `git` repository for installation of their plugins:
```toml
[plugins.neorg]
git = "nvim-neorg/neorg"

# or...

[plugins]
neorg = { git = "nvim-neorg/neorg", version = "1.0.0" }
```

linrongbin16:
this feature would definitely help more users migrate to this plugin manager.

and also support git tags/branches!

mrcjkb:
Some things to consider:

- Do we want to use [packages](https://neovim.io/doc/user/usr_05.html#add-package) (type = `opt` | `start`, like minpac, paq, ...)?
- Or should we use a rockspec template to generate a simple rockspec (like we do with luarocks-tag-release) and hook into the rocks install?
- Or both (type = `opt` | `start` | `rock`)?

Personally, I like the idea of generating a rockspec.

mrcjkb:
> and how does this module behave when there are two plugins depend on a same git repository, but with different versions/tags/branches?

I think handling dependencies would be outside the scope of this module.

The point of rocks.nvim is to take away that burden from the user and make it the plugin authors' responsibility.
So we just need to rigourously package plugins that have dependencies for luarocks, so they can be installed without the git module

mrcjkb:
> can rocks.nvim support isolated versions on the same dependency?

luarocks can install different versions of the same package, but I don't know at this point if it causes issues at runtime. My guess would be that a `require` will load the first module it finds on the path.

We should probably include some kind of dependency conflict detection at least in the health checks.

linrongbin16:
> > can rocks.nvim support isolated versions on the same dependency?
> 
> luarocks can install different versions of the same package, but I don't know at this point if it causes issues at runtime. My guess would be that a `require` will load the first module it finds on the path.
> 
> We should probably include some kind of dependency conflict detection at least in the health checks.

maybe first investigate if isolated runtime path is supported for a supported by Neovim.

mrcjkb:
We don't add dependencies to the rtp, just to the `package.path` and `package.cpath`.

But [it looks like](https://github.com/Kong/kong/issues/1761) luarocks handles conflicting dependencies as it should.

linrongbin16:
> We don't add dependencies to the rtp, just to the `package.path` and `package.cpath`.
> 
[...]

PR #85, opened 2023-12-17
feat(api): allow external modules to hook into `:Rocks sync`

mrcjkb:
... in preparation for #81.

This is ready to be reviewed, but I'd prefer to leave it as a draft and not to merge it until we have a basic working implementation of a module that uses this API.

I initially wanted the API to allow hooking into the `update`, `sync` and `prune` commands.
But after playing around with the idea, I think `sync` is enough. 

Hooking into `update` and `prune` might be a bad idea, for the following reasons:

- `update` uses the luarocks state to determine the outdated rocks, and then updates the rocks.toml (parsed with `toml-edit`).
   Hooking into `update` would require parsing the config with `toml`, and potentially allowing to update the config with `toml-edit`.
   This adds a whole lot of complexity. 
- Like `update`, `prune` parses and modifies the config with `toml-edit`. Hooking into it would add complexity.
- The `rocks-git` module is meant to be a bridge for making rocks.nvim adoption less painful while there aren't many plugins on luarocks. I don't think it's desirable to implement features like globs on tags (people can use Lazy for that).
  If we do end up needing those features, it would be better for the module to implement its own commands (e.g. `:Rocks gitPull`).

In terms of the `rocks-git` module, we can treat `sync` as follows:

- Pull if there is no `rev` or `tag` field, as this typically means the user wants the `HEAD` of the main branch.
- Checkout a `rev` or `tag` if one is specified.

### DIFF
--- a/doc/rocks.txt
+++ b/doc/rocks.txt
@@ -5,6 +5,7 @@ rocks.nvim ··································�
 rocks.nvim commands ··········································· |rocks.commands|
 rocks.nvim configuration ········································ |rocks.config|
 rocks.nvim Lua API ················································· |rocks.api|
+rocks.nvim logging API ············································· |rocks.log|
 
 ==============================================================================
 rocks.nvim                                                               *rocks*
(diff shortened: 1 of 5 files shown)


vhyrro:
Design decisions sound good to me thus far :)

I'll review fully once this stops being a draft PR (aka after rocks-git sees its first release).

vhyrro (approved):

PR #90, opened 2023-12-17
fix(sync): Better error message if entry can't be parsed

mrcjkb:


lua/rocks/operations.lua
```diff
@@ -305,6 +305,12 @@ operations.sync = function(user_rocks)
 
         for _, key in ipairs(to_install) do
             nio.scheduler()
+            if not user_rocks[key].version then
+                local message = ("Could not parse rock: %s"):format(vim.inspect(user_rocks[key]))
+                log.error(message)
+                report_error(message)
+                goto skip_install
```
NTBBloodbath, line 312: question (non-blocking): this will not work with Neovim builds that are not using LuaJIT. I'm using a `goto` statement in `rest.nvim` as well and I haven't seen any problems since that is an edge case of a custom Neovim build but it's still good to clarify it. Should something regarding this be added to the README?

mrcjkb: Let's just avoid `goto` then :smile:

NTBBloodbath: If we have an alternative or we want to maintain compatibility with Neovim builds not compiled with LuaJIT then yes. For `rest.nvim` as I said, I haven't seen anyone opening an issue due to incompatibilities in more or less two years.

The only time I had problems with that was when Gentoo was building Neovim without LuaJIT support yet, and that was when version 0.5 was the newest, so I guess it's fine to use it :p

mrcjkb: Ah, then let's leave it.
It does save us an indentation level.

PR #104, opened 2023-12-29
fix: better stack traces

mrcjkb:
`pcall` unwinds the stack when it fails, so we lose the stack trace. Moving the code to a different
coroutine allows us to notify/log the stack trace
that we're actually interested in.

### DIFF
--- a/lua/rocks/operations.lua
+++ b/lua/rocks/operations.lua
@@ -482,28 +482,32 @@ operations.update = function()
                 name = name,
                 version = rock.target_version,
             })
-            local success, ret = pcall(future.wait)
+            local co = coroutine.create(function()
+                coroutine.yield(future.wait())
+            end)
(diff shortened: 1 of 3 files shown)


lua/rocks/operations.lua
```diff
@@ -482,20 +482,22 @@ operations.update = function()
...
                 progress_handle:report({
                     percentage = get_percentage(ct, #outdated_rocks),
                 })
             end
-            user_rocks.plugins[ret.name] = ret.version
-            progress_handle:report({
-                message = ("Updated %s: %s -> %s"):format(rock.name, rock.version, rock.target_version),
-                percentage = get_percentage(ct, #outdated_rocks),
-            })
         end
```
mrcjkb, line 501: This was a bug I fixed on the fly - we shouldn't be attempting to index `[ret.name]` if the function that produces `ret` failed.

PR #112, opened 2024-01-14
fix(runtime): variable naming conflict

### LINKED ISSUE #110: Attempt to call local 'type' (a string value)
```
Error detected while processing /home/xiaoshihou/.local/share/nvim/rocks/lib/luarocks/rocks-5.1/rocks.nvim/2.6.1-1/plugin/rocks.lua:               
E5113: Error while calling lua chunk: .../.local/share/nvim/rocks/share/lua/5.1/rocks/runtime.lua:72: attempt to call local 'type' (a string value)
```

Maybe it's a good idea to rename the variable `type`
https://github.com/nvim-neorocks/rocks.nvim/blob/7d786a28060fa35f344b2df53c01ab12090b7133/lua/rocks/runtime.lua#L68

mrcjkb:
Fixes #110.

### DIFF
--- a/lua/rocks/runtime.lua
+++ b/lua/rocks/runtime.lua
@@ -46,17 +46,17 @@ local function iter_children(dir)
     return coroutine.wrap(function()
         local handle = vim.uv.fs_scandir(dir)
         while handle do
-            local name, type = vim.uv.fs_scandir_next(handle)
+            local name, ty = vim.uv.fs_scandir_next(handle)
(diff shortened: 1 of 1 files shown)


mrcjkb:
Merging without review, as this is trivial.

NTBBloodbath (changes requested):
Hmm why not call it `_type` or `type_` instead? I think it makes more sense for readability and doesn't collide with the built-in `type` 🤔

PR #116, opened 2024-01-17
feat: settings.lazy for disabling plugin loading

xiaoshihou514:
Problem: Sometimes one wants *non* of the plugins to be sourced on `rocks.nvim` startup, but adding `opt = false` for all the plugins is tiresome

Solution: introduces a `settings` header, with a `lazy` field, if that is true, we wouldn't source any plugins on startup

What do you guys think 🤔 ?

### DIFF
--- a/lua/rocks/config/internal.lua
+++ b/lua/rocks/config/internal.lua
@@ -48,7 +48,7 @@ local default_config = {
         local config_file = fs.read_or_create(config.config_path, constants.DEFAULT_CONFIG)
         local rocks_toml = require("toml").decode(config_file)
         for key, tbl in pairs(rocks_toml) do
-            if key == "rocks" or key == "plugins" then
+            if key == "rocks" or key == "plugins" or key == "settings" then
                 for name, data in pairs(tbl) do
                     if type(data) == "string" then
                         ---@type RockSpec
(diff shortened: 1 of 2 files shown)


mrcjkb:
Hey :wave: 

Thanks for the PR!
I'm not quite sure how I feel about this:

- It can be a footgun for newcomers.
  For example, I've had lazy.nvim users who have that option set in their configs open issues, because they didn't understand the implications, and thought my plugins weren't working.
- This doesn't propagate to external modules, like `rocks-git.nvim` - although that could be implemented in the module.

Personally, I think a good compromise would be to extend the `:Rocks install` command to take an optional `opt=true` or  `opt=1`  argument (or just `opt`?) .

@vhyrro @teto @NTBBloodbath what are your thoughts?

xiaoshihou514:
@mrcjkb You are totally right that this might be confusing. Would more explicit naming, like `load_plugins = "false"` address that?

mrcjkb:
I'm not sure how a settings header would be handled by `rocks-git.nvim`, because it interests `opt=true` as "install this plugin into the `opt` packpath". It has no concept of "loading plugins".

There's also the question of what to do if `load_plugins = false`, but one or more plugins have explicitly set `opt = false`?

This increases the complexity in a way I don't think I feel comfortable with.
But I'll wait for feedback from the others 😃

teto:
I dont have a good grasp about the problem but if you are not comfortable with it, I would say let's not.
rocks.nvim introduces a different kind of complexity compared to the traditional approach. Let's focus on making this robust and we can add a `disabled` thing later on. Lazyness is overrated anyway

xiaoshihou514:
I see, I guess it was the part of me that cares about every ms being paranoid.

PR #129, opened 2024-01-27
docs(readme): add suggestion to run installer with `-u NORC`

mrcjkb:


NTBBloodbath (commented):
Should we also mention that it's because some plugins can interfere with the installer script?

mrcjkb:
> Should we also mention that it's because some plugins can interfere with the installer script?

Done

NTBBloodbath (approved):

PR #130, opened 2024-02-01
fix(installer): remove color seqence from shell output

pysan3:
## Lua is not oxidized

https://github.com/nvim-neorocks/rocks.nvim/blob/4f9086551bd9af0eeead663ebed54eaddf67f1e3/installer.lua#L170

Lua has no mustache.
Replace all with `%s`.

## Remove color sequences

But even if I replace all with `%s`, the error message is pretty difficult to read due to the color sequences in the output.

image

I've added a wrapper function to remove the color sequences for a better visibility.

### DIFF
--- a/installer.lua
+++ b/installer.lua
@@ -137,6 +137,24 @@ local function guard_set_up_luarocks_dependency_missing(dep)
     return false
 end
 
+--- Notify command output.
+---@param msg string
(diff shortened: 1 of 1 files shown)


mrcjkb (approved):
This is awesome! Thanks :pray:

PR #143, opened 2024-02-11
fix(update): don't add dependencies to rocks.toml after updating

mrcjkb:


mrcjkb:
Merging, as this is trivial...

PR #145, opened 2024-02-12
feat: generate plugins help pages tags on install/update

NTBBloodbath:
This PR also adds a new configuration option called `generate_help_pages`, which defaults to `true`.

### DIFF
--- a/lua/rocks/config/internal.lua
+++ b/lua/rocks/config/internal.lua
@@ -36,6 +36,8 @@ local default_config = {
     lazy = false,
     ---@type boolean Whether to automatically add freshly installed plugins to the 'runtimepath'
     dynamic_rtp = true,
+    ---@type boolean Whether to re-generate plugins help pages after installation/upgrade
+    generate_help_pages = true,
     ---@class RocksConfigDebugInfo
     debug_info = {
         ---@type boolean
--- a/lua/rocks/operations.lua
+++ b/lua/rocks/operations.lua
@@ -127,7 +127,7 @@ operations.install = function(rock_spec, progress_handle)
 
             -- Re-generate help tags
             if config.generate_help_pages then
-                vim.cmd("heltags ALL")
+                vim.cmd("helptags ALL")
             end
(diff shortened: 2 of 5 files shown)


NTBBloodbath:
Hmm I did a `push --force` because I had committed a typo before in a command and Git doesn't seem to care about the flag. It would be nice to do a squash

lua/rocks/operations.lua
```diff
@@ -125,6 +125,11 @@ operations.install = function(rock_spec, progress_handle)
                 runtime.packadd(name)
             end
 
+            -- Re-generate help tags
+            if config.generate_help_pages then
+                vim.cmd("helptags ALL")
+            end
+
```
mrcjkb, line 132: question: Wouldn't it make more sense to put this at the end of `operations.add` and `operations.sync`, instead of in `operations.install`?

That way, we only call it once, rather than multiple times for each plugin that is added, e.g. in `sync`.

lua/rocks/config/internal.lua
```diff
@@ -36,6 +36,8 @@ local default_config = {
     lazy = false,
     ---@type boolean Whether to automatically add freshly installed plugins to the 'runtimepath'
     dynamic_rtp = true,
+    ---@type boolean Whether to re-generate plugins help pages after installation/upgrade
+    generate_help_pages = true,
```
mrcjkb, line 40: todo: we should add this to the user-facing `config` module (`config/init.lua`), too (CI will generate vimdoc from it).

mrcjkb (changes requested):

lua/rocks/config/internal.lua
NTBBloodbath: Woops, I don't know how I've missed this one, thank you!

vhyrro (approved):

mrcjkb (approved):

mrcjkb:
Weird that this isn't triggering CI...

PR #150, opened 2024-02-16
fix(luarocks): make sure `LUAROCKS_CONFIG` is unset

### LINKED ISSUE #149: [Bug] If `LUAROCKS_CONFIG` is set, it will override rocks.nvim's luarocks config
Ton option to mitigate this would be to

- get the environment variable before running any luarocks commands
- unset it
- run the luarocks command
- reset it

mrcjkb:
Fixes #149


mrcjkb:
Merging, as this is trivial.

PR #153, opened 2024-02-17
feat: add luarocks `bin` directory to the Neovim path

vhyrro:
This PR allows any plugin running under a Neovim instance with `rocks.nvim` to gain access to whatever binaries are installed under `.local/share/nvim/rocks/bin/`. This may come in super handy for rocks which publish binaries ([wink wink](https://github.com/nvim-neorg/norgopolis-server.lua)) :)

### DIFF
--- a/lua/rocks/config/internal.lua
+++ b/lua/rocks/config/internal.lua
@@ -97,10 +97,28 @@ if #config.debug_info.unrecognized_configs > 0 then
     )
 end
 
+---@type
+---|"sysv"
+---| "hpux"
+---| "netbsd"
+---| "linux"
+---| "hurd"
+---| "solaris"
+---| "aix"
+---| "irix"
+---| "freebsd"
+---| "openbsd"
+---| "dragonfly"
+---| "haiku"
+---| "windows"
+---| "cygwin"
+---| "macosx"
+local current_os = assert((require("sysdetect").detect()), "[rocks]: unable to detect current operating system!")
+
+local env_path_seperator = current_os == "windows" and ";" or ":"
+
 -- Append the binary directory to the system path.
--- Note: All major operating system follow the same PATH structure, so there should be little
--- issue with modifying the environment variable this way.
-vim.env.PATH = vim.fs.joinpath(config.rocks_path, "bin") .. ":" .. vim.env.PATH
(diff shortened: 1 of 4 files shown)


vhyrro:
I found the end of config.lua to the best place to put the path expansion logic. Two questions:

- We should probably prepend the bin path instead of append to give it a higher priority, right?
- Should I move it from config.lua to somewhere else?

(cc @mrcjkb)

vhyrro:
Nevermind, for some reason I thought Windows also uses `:` as its path separator. It actually uses a semicolon - thanks, windows.

vhyrro:
Added hisham's very recent `sysdetect` rock to rocks. This will ensure that we're using the correct path separator everywhere hopefully :)

vhyrro:
image
Hmm...

mrcjkb:
> image Hmm...

The nix build is sandboxed. We'll probably have to package sysdetect and add it to the nix build inputs.
I'll look into it later today

mrcjkb:
> I found the end of config.lua to the best place to put the path expansion logic. Two questions:
> 
> * ~We should probably prepend the bin path instead of append to give it a higher priority, right?~ Done that.
> * Should I move it from config.lua to somewhere else?
> 
> (cc @mrcjkb)

How about `plugin/rocks.lua`?

lua/rocks/config/internal.lua
```diff
@@ -97,6 +97,29 @@ if #config.debug_info.unrecognized_configs > 0 then
...
+---| "aix"
+---| "irix"
+---| "freebsd"
+---| "openbsd"
+---| "dragonfly"
+---| "haiku"
+---| "windows"
+---| "cygwin"
+---| "macosx"
+local current_os = assert((require("sysdetect").detect()), "[rocks]: unable to detect current operating system!")
```
mrcjkb, line 116: We could also check if  `vim.uv.os_uname().sysname`  is `"Windows"`.

vhyrro: I did consider that but I thought that being open to more accurate system detection wouldn't hurt (just in case some random OS decides to have a different PATH separator lol). This rock can also detect architecture which we may (?) need.

mrcjkb: I would say YAGNI for now. But it's good to know about the package in case we end up needing it 😀

vhyrro:
Fixed now!

mrcjkb (approved):

PR #165, opened 2024-02-26
feat(api): ability to hook into `:Rocks install` and `:Rocks update`

mrcjkb:


mrcjkb:
Marking this as draft, because I still need to implement the rocks-git side and test it.

NTBBloodbath (approved):

PR #184, opened 2024-03-13
feat: add support for tree-sitter parsers installed from luarocks :deciduous_tree:

mrcjkb:


NTBBloodbath (approved):
LGTM, thank you very much for all the hard work on this front mate :)

PR #186, opened 2024-03-13
fix(installer): proper C lib file extension on darwin and windows

mrcjkb:


mrcjkb:
This one is trivial. Merging.

Issue #166, opened 2024-02-27
Windows support


tembokk:
My nvim version: `v0.10.0-dev-2388+gb12d193b4` 
I install `lua` and `luarocks` through `scoop install lua-for-windows`
Then on rocks.nvim installation I set `Set up luarocks (recommended) ?:` **`false`**
And I get this error 
```
Installing rocks.nvim...
E5108: Error executing lua: ...m-nightly\current\share\nvim\runtime/lua/vim/_system.lua:241: ENOENT: no such file or directory
stack traceback:
	[C]: in function 'error'
	...m-nightly\current\share\nvim\runtime/lua/vim/_system.lua:241: in function 'spawn'
	...m-nightly\current\share\nvim\runtime/lua/vim/_system.lua:332: in function 'system'
	C:\Users\Windows 10\AppData\Local\Temp\nvim.0\ywhunR\0.lua:383: in function <C:\Users\Windows 10\AppData\Local\Temp\nvim.0\ywhunR\0.lua:357>
```

mrcjkb:
Which version of luarocks does scoop install and what's the output of `luarocks --version` on the command line?

mrcjkb:
Thanks for troubleshooting this!

I've looked at the neovim source. `vim.system` uses `uv.spawn` under the hood, which says:

>  Possible reasons for failing to spawn would include (but not be limited to) the
> file to execute not existing, not having permissions to use the setuid or setgid
> specified, or not having enough memory to allocate for the new process.

and can confirm that `ENOENT: no such file or directory` is thrown when you pass in a binary it can't find.

Regarding: `Lua 5.1 interpreter not found in $PATH`:

We use `sh configure` to configure luarocks. My guess is that it falls back to `git sh` on Windows (I don't think `sh` is installed by default, but git for Windows installs it iirc?). So it's likely that `git sh` doesn't see the `PATH`.

@vhyrro What do you think? Maybe we should use `vim.o.sh` for this?
@tembokk What's the output of `:lua =vim.o.sh` on your Windows machine?

luarocks 2.0.2 is very outdated and unlikely to work with rocks.nvim (we should probably add a minimum version requirement check to the installer).

> `Error: Failed finding Lua header files. You may need to install them or configure LUA_INCDIR.`

Are you still using lua installed with `scoop`? luarocks makes some assumptions about where to find header files.

[...]

PR #200, opened 2024-03-17
revert(installer): use `vim.o.sh` instead of `'sh'`

### LINKED ISSUE #199: Fails in fish shell
```
Configuring luarocks failed.                                                                                  
stderr: configure (line 5): Unsupported use of '='. In fish, please use 'set prefix "/usr/local"'.            
prefix="/usr/local"                                                                                           
^~~~~~~~~~~~~~~~~~^                                                                                           
warning: Error while reading file configure                                                                   
                                                                                                              
                           
```

gptlang:
Fixes #199 

Tested on Fedora 39 with `NVIM v0.10.0-dev+2606-gc0549b9c4`

### DIFF
--- a/installer.lua
+++ b/installer.lua
@@ -185,8 +185,7 @@ local function set_up_luarocks(install_path)
     vim.notify("Configuring luarocks...")
 
     sc = vim.system({
-        vim.o.sh,
-        "configure",
+        "./configure",
         "--prefix=" .. install_path,
         "--lua-version=5.1",
         "--force-config",
--- a/installer.lua
+++ b/installer.lua
@@ -185,7 +185,8 @@ local function set_up_luarocks(install_path)
     vim.notify("Configuring luarocks...")
 
     sc = vim.system({
-        "./configure",
+        "sh"
+        "configure",
         "--prefix=" .. install_path,
         "--lua-version=5.1",
         "--force-config",


mrcjkb (approved):
Thanks.

This breaks the installer for Windows (see #166).
But using the configure script on Windows might not be the best idea anyway.

mrcjkb:
I'll try to come up with a proper solution for Windows today, so I'll wait with merging for now.

vhyrro:
This might be a risky thing to attempt, as we can't confirm that the binary will be executable on the target filesystem. When I tested (Arch), I get a "configure is not executable" error. I do think that we should use `sh` instead of `vim.o.sh` however to make sure we're running a POSIX-compatible shell :)

gptlang:
> This might be a risky thing to attempt, as we can't confirm that the binary will be executable on the target filesystem. When I tested (Arch), I get a "configure is not executable" error. I do think that we should use `sh` instead of `vim.o.sh` however to make sure we're running a POSIX-compatible shell :)

Yeah that makes sense to me. Would it make sense to hard code `/bin/sh` or `/bin/env sh`?

mrcjkb:
> > This might be a risky thing to attempt, as we can't confirm that the binary will be executable on the target filesystem. When I tested (Arch), I get a "configure is not executable" error. I do think that we should use `sh` instead of `vim.o.sh` however to make sure we're running a POSIX-compatible shell :)
> 
> Yeah that makes sense to me. Would it make sense to hard code `/bin/sh` or `/bin/env sh`?

I believe `/bin/sh` (or just `sh`) is fine. `/usr/bin/env` is more commonly used for `bash` (which is less commonly available in `/bin`.

NTBBloodbath:
> > > This might be a risky thing to attempt, as we can't confirm that the binary will be executable on the target filesystem. When I tested (Arch), I get a "configure is not executable" error. I do think that we should use `sh` instead of `vim.o.sh` however to make sure we're running a POSIX-compatible shell :)
> > 
> > Yeah that makes sense to me. Would it make sense to hard code `/bin/sh` or `/bin/env sh`?
> 
> I believe `/bin/sh` (or just `sh`) is fine. `/usr/bin/env` is more commonly used for `bash` (which is less commonly available in `/bin`.

I think using `/usr/bin/env sh` would be best. It is the best way to maintain compatibility with any shell even in scripts for BSD systems iirc, I would go with `/bin/sh` but I think that would break on certain systems like Alpine, which is an edge case but it's good to keep it in mind

mrcjkb:
> > > > This might be a risky thing to attempt, as we can't confirm that the binary will be executable on the target filesystem. When I tested (Arch), I get a "configure is not executable" error. I do think that we should use `sh` instead of `vim.o.sh` however to make sure we're running a POSIX-compatible shell :)
> > > 
> > > 
> > > Yeah that makes sense to me. Would it make sense to hard code `/bin/sh` or `/bin/env sh`?
> > 
> > 
> > I believe `/bin/sh` (or just `sh`) is fine. `/usr/bin/env` is more commonly used for `bash` (which is less commonly available in `/bin`.
> 
> I think using `/usr/bin/env sh` would be best. It is the best way to maintain compatibility with any shell even in scripts for BSD systems iirc, I would go with `/bin/sh` but I think that would break on certain systems like Alpine, which is an edge case but it's good to keep it in mind

That only matters with `bash`, not `sh`. 

https://stackoverflow.com/questions/53116226/what-is-the-recommended-posix-sh-shebang

mrcjkb (approved):

mrcjkb:
Reverts https://github.com/nvim-neorocks/rocks.nvim/commit/54b67cec53f4152c85fb8f7b77a939514b5f78a5

PR #211, opened 2024-03-19
feat(sync): bootstrap external sync handlers and use them in one sync…

mrcjkb:
… sweep


mrcjkb:
Merging, as this is quite small and it works like a charm :rocket:

PR #217, opened 2024-03-21
feat(api): `install` function for installing rocks with a callback

mrcjkb:


mrcjkb:
Merging, as this is quite trivial

PR #227, opened 2024-03-26
feat(loader): support multiple versions of the same dependency

### LINKED ISSUE #215: [Feature] support multiple versions of the same dependency
This should be possible by setting up the [luarocks loader](https://github.com/luarocks/luarocks/wiki/Using-LuaRocks#multiple-versions-using-the-luarocks-package-loader).

We may need to provide a `LUAROCKS_CONFIG` file and a way to get the path to the loader in the luarocks installation.

mrcjkb:
Closes #215

### DIFF
--- a/plugin/rocks.lua
+++ b/plugin/rocks.lua
@@ -6,6 +6,30 @@ local nio = require("nio")
 local adapter = require("rocks.adapter")
 local config = require("rocks.config.internal")
 
+local function get_luarocks_loader_path_from_luarocks()
+    local sc = vim.system({ config.luarocks_binary, "which", "luarocks.loader" }):wait()
+    return sc.stdout and sc.stdout:match("(%S+)loader.lua")
+end
+
+-- Initialize the luarocks loader
+if config.enable_luarocks_loader then
+    local default_luarocks_binary = vim.fs.joinpath(config.rocks_path, "bin", "luarocks")
+    local luarocks_loader_path = config.luarocks_binary == default_luarocks_binary
+            and vim.fs.joinpath(default_luarocks_binary, "share", "lua", "5.1", "luarocks", "?.lua")
+        or get_luarocks_loader_path_from_luarocks()
+    if luarocks_loader_path then
+        package.path = package.path .. ";" .. luarocks_loader_path .. "?.lua"
+        vim.env.LUAROCKS_CONFIG = config.luarocks_config
+        local ok, err = pcall(require, "luarocks.loader")
+        -- TODO: log errors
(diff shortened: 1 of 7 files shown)


plugin/rocks.lua
```diff
@@ -6,6 +6,25 @@ local nio = require("nio")
 local adapter = require("rocks.adapter")
 local config = require("rocks.config.internal")
 
+-- Initialize the luarocks loader
+if config.enable_luarocks_loader then
+    local sc = vim.system({ config.luarocks_binary, "which", "luarocks.loader" }):wait()
+    local luarocks_path = sc.stdout and sc.stdout:match("(%S+)loader.lua")
```
vhyrro, line 12: I assume we are doing the whole "luarocks which" check as we also permit users to use their system installation of luarocks, right?

plugin/rocks.lua
```diff
@@ -6,6 +6,25 @@ local nio = require("nio")
...
+    if luarocks_path then
+        package.path = package.path .. ";" .. luarocks_path .. "?.lua"
+        vim.uv.os_setenv("LUAROCKS_CONFIG", config.luarocks_config)
+        local ok, err = pcall(function()
+            require("luarocks.loader")
+        end)
```
vhyrro, line 18:
```suggestion
        local ok, err = pcall(require, "luarocks.loader")
```

This bit can be shortened :)

plugin/rocks.lua
```diff
@@ -6,6 +6,25 @@ local nio = require("nio")
 local adapter = require("rocks.adapter")
 local config = require("rocks.config.internal")
 
+-- Initialize the luarocks loader
+if config.enable_luarocks_loader then
+    local sc = vim.system({ config.luarocks_binary, "which", "luarocks.loader" }):wait()
+    local luarocks_path = sc.stdout and sc.stdout:match("(%S+)loader.lua")
+    if luarocks_path then
+        package.path = package.path .. ";" .. luarocks_path .. "?.lua"
+        vim.uv.os_setenv("LUAROCKS_CONFIG", config.luarocks_config)
```
vhyrro, line 15:
```suggestion
        vim.env.LUAROCKS_CONFIG = config.luarocks_config
```

nit: would this not be the simpler approach, or does it have some drawbacks I'm not aware of? :)

plugin/rocks.lua
mrcjkb: Yes.
I've just added a change that doesn't call `vim.system` if the luarocks path `<rocks_config>/bin/luarocks`.

PR #236, opened 2024-04-02
feat(install): support passing args, like `opt=true`

mrcjkb:
This allows the following commands:

```vim
:Rocks install neorg
:Rocks install neorg 8.0.0
:Rocks install neorg 8.0.0 opt=true
:Rocks install neorg opt=true 8.0.0
:Rocks install neorg opt=true version=8.0.0
:Rocks install neorg opt=true
...
```

### DIFF
--- a/lua/rocks/operations/init.lua
+++ b/lua/rocks/operations/init.lua
@@ -522,7 +522,7 @@ operations.add = function(arg_list, callback)
         end
         ---@type rock_name
         local rock_name = arg_list[1]
-        --- We can't mutate the arg_list, because we may need it for a recursive add
+        -- We can't mutate the arg_list, because we may need it for a recursive add
         ---@type string[]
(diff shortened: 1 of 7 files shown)


lua/rocks/operations/init.lua
```diff
@@ -499,20 +513,29 @@ operations.add = function(arg_list, rock_name, version, callback)
...
             handler(report_progress, report_error)
             vim_schedule_nio_wait(function()
                 fs.write_file(config.config_path, "w", tostring(user_rocks))
                 progress_handle:finish()
             end)
             return
         end
+        ---@type rock_name
+        local rock_name = arg_list[1]
+        --- We can't mutate the arg_list, because we may need it for a recursive add
```
vhyrro, line 525:
```suggestion
        -- We can't mutate the arg_list, because we may need it for a recursive add
```

nitpick: probably shouldn't be a docstring :p

vhyrro:
Other than the minor nitpick LGTM!

PR #244, opened 2024-04-05
feat: lockfile support

### LINKED ISSUE #34: Feature request: support pinning
I think it would be very useful to support a lockfile to always install the plugins and their dependencies with the same version if you have to change the Neovim installation either because you installed your dotfiles on another machine or another reason.

Luarocks supports this natively, please take a look at [Pinning versions with a lock file](https://github.com/luarocks/luarocks/wiki/Pinning-versions-with-a-lock-file).

mrcjkb:
Closes #34.

The [documentation](https://github.com/luarocks/luarocks/wiki/Pinning-versions-with-a-lock-file#using-pinned-dependencies-in-luarocks) says luarocks will use a lockfile if there is one in the cwd.
But I haven't confirmed this yet.

This needs some vigorous testing.

### How it works

Adding/updating/removing rocks: 

- We add a `--pin` flag to the `install` command
- luarocks creates a luarocks.lock file in the rock's install tree (for all transitive dependencies of the installed rock)
- We add the path to the rtp
- We load the lockfile and use it to update our rocks-lock.toml

Syncing rocks:

- We get the locked dependencies from our rocks-lock.toml
- We use those to create a luarocks.lock in a temp directory
- We set the `cwd` to the location of the lock file for any `install` commands

### DIFF
--- a/lua/rocks/operations/lock.lua
+++ b/lua/rocks/operations/lock.lua
@@ -1,0 +1,87 @@
+---@mod rocks.operations.lock
+--
+-- Copyright (C) 2024 Neorocks Org.
+--
+-- License:    GPLv3
+-- Created:    7 Jul 2024
+-- Updated:    7 Jul 2024
+-- Homepage:   https://github.com/nvim-neorocks/rocks.nvim
+-- Maintainers: NTBBloodbath <bloodbathalchemist@protonmail.com>, Vhyrro <vhyrro@gmail.com>, mrcjkb <marc@jakobi.dev>
+--
+---@brief [[
+--
+-- Lockfile management.
+--
+---@brief ]]
+
+local config = require("rocks.config.internal")
+local fs = require("rocks.fs")
+local nio = require("nio")
+
+local lock = {}
+
+---@param reset boolean
+local function parse_rocks_lock(reset)
+    local lockfile = reset and "" or fs.read_or_create(config.lockfile_path, "")
+    return require("toml_edit").parse(lockfile)
+end
+
+---@param rock_name? rock_name
+lock.update_lockfile = nio.create(function(rock_name)
+    local luarocks_lockfiles = vim.iter(vim.api.nvim_get_runtime_file("luarocks.lock", true))
+        :filter(function(path)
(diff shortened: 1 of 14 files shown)


mrcjkb:
:thinking: maybe we should just call the file `rocks.lock` instead of `rocks-lock.toml`.

lua/rocks/operations/lock.lua
```diff
@@ -0,0 +1,78 @@
...
+            if not ok or not loader then
+                return
+            end
+            local success, luarocks_lock_tbl = pcall(loader)
+            if not success or not luarocks_lock_tbl or not luarocks_lock_tbl.dependencies then
+                return
+            end
+            rocks_lock[rock_key] = {}
+            for dep, version in pairs(luarocks_lock_tbl.dependencies) do
+                if dep ~= "lua" then
```
mrcjkb, line 50: I think we should also exclude non semver versions from the lockfile.

Pinning `scm` could lead to subtle bugs. Perhaps we should also create a luarocks issue.

mrcjkb:
I just can't get the test cases to succeed (╯°□°)╯︵ ┻━┻

It looks like it could be something to do with the neovim scheduler? Not sure how to get busted working with that.

mrcjkb:
This PR is pretty much ready. But I would like to merge #245 first.

mrcjkb:
Todo:

- [ ] Update the new `sync_spec.lua` to install a package that has a locked dependency, and confirm that the locked dependency is indeed installed.

mrcjkb:
Looks like luarocks is ignoring the lockfile...

mrcjkb:
Blocked by https://github.com/luarocks/luarocks/issues/1662

purepani:
Looks like https://github.com/luarocks/luarocks/issues/1662 has been fixed, so this is now unblocked

mrcjkb:
> Looks like [luarocks/luarocks#1662](https://github.com/luarocks/luarocks/issues/1662) has been fixed, so this is now unblocked

We're going to have to wait for luarocks 3.11.1 to land in nixpkgs.

mrcjkb:
Well... it looks like luarocks 3.11.1 is still not picking up the lockfile when installing :disappointed:

PR #259, opened 2024-04-07
build(nix): use nix-buildproxy to sandbox tests

mrcjkb:


mrcjkb:
I don't have it working yet, and I'm not sure if I can get it to work with luarocks. Might have to pick it up another time.

Apparently, luarocks [supports `HTTP_PROXY`](https://github.com/luarocks/luarocks/wiki/LuaRocks-through-a-proxy).

PR #281, opened 2024-04-18
[Feature]: Add bootstrapping support.

vhyrro:
Apart from an installer one may want to install rocks.nvim through a bootstrapping script.

This PR adds support for such a bootstrapping script for installing rocks.nvim whenever it can't be found.

#### Notes:
- [ ] We may want to mention that *order matters* in the bootstrapping script. I.e. - the rtp should be modified before the existence of rocks.nvim is checked for.
- [ ] The readme shouldn't point to the pull request branch, that was just for testing :sweat_smile:


README.md
```diff
@@ -219,6 +219,73 @@ nvim -u NORC -c "source https://raw.githubusercontent.com/nvim-neorocks/rocks.nv
...
+            "git",
+            "clone",
+            "--filter=blob:none",
+            "https://github.com/nvim-neorocks/rocks.nvim",
+            rocks_location,
+        })
+    end
+
+    -- If the clone was successful then source the bootstrapping script
+    assert(vim.v.shell_error == 0, "rocks.nvim installation failed. Try exiting and re-entering Neovim!")
```
NTBBloodbath, line 277: nitpick (non-blocking): perhaps we could use `vim.system` API instead as we are already using Neovim nightly APIs?

bootstrap.lua
```diff
@@ -0,0 +1,125 @@
...
+-- - Use the new luarocks installation to install `rocks.nvim`
+
+-- The rocks.nvim plugin is already loaded via the vim.opt.runtimepath:append()
+-- call in the `init.lua` bootstrapping script.
+
+math.randomseed(os.time())
+
+local config_data = vim.g.rocks_nvim or {}
+local install_path = config_data.rocks_path or vim.fs.joinpath(vim.fn.stdpath("data"), "rocks")
+local luarocks_binary = config_data.luarocks_binary or vim.fs.joinpath(install_location, "bin", "luarocks")
```
NTBBloodbath, line 14:
```suggestion
local luarocks_binary = config_data.luarocks_binary or vim.fs.joinpath(install_path, "bin", "luarocks")
```

bootstrap.lua
```diff
@@ -0,0 +1,125 @@
...
+            sc and "stdout: " .. remove_shell_color(sc.stdout),
+        }, "\n"),
+        level
+    )
+end
+
+--- Sets up luarocks for use with rocks.nvim
+---@param install_path string
+---@return boolean success
+local function set_up_luarocks(install_path)
```
NTBBloodbath, line 47: issue: this parameter needs to be renamed, as it collides with the top-level variable `install_path` and causes `luacheck` to fail in the CI due to shadowing value.

NTBBloodbath (changes requested):

mrcjkb:
suggestion(non-blocking): It would be nice to have an integration test for this (not using nix - I don't think it would be possible to sandbox it). It could be a matrix over all platforms, replacing the existing Windows test.

- init.lua with the bootstrapping logic
  - Maybe a CI action that injects the init.lua into the readme so we have a single source of truth? Or the other way around?
- Start neovim
- Try to run a `:Rocks` command

Feel free to create an issue from this comment if you don't want to do it with this PR.

README.md
vhyrro: I chose to keep vim.fn.system just because vim.v.shell_error is nicer to deal with in this specific scenario. With vim.system we'd have to capture the output into a variable which would have to have a greater scope than the if statement -- too much haslle really :p

vhyrro:
@NTBBloodbath changes have been addressed, let me know if there's anything else to do. In the meantime I'll work on addressing the first TODO in the PR message :)

mrcjkb (approved):

PR #306, opened 2024-04-29
fix(luarocks): add `--force-lock` flag

mrcjkb:
If exiting neovim duringa `:Rocks sync`, a lockfile.lfs may be left behind in the rocks_dir.
We already have an in-memory locking mechanism, so there's no need to use the luarocks lock file, which currently needs to be removed manually.


mrcjkb:
Merging, as this is trivial.

PR #323, opened 2024-05-08
fix: wait for writing rocks.toml to complete where possible

mrcjkb:


mrcjkb:
CI failure is fixed in #324

PR #340, opened 2024-05-17
feat: manage luarocks installation as rockspec dependency

### LINKED ISSUE #336: [Enhancement] pin luarocks version / manage luarocks installation
Our installer currently builds luarocks from source (`HEAD`), which could be bad if something breaks.
It would be better for us to pin the version.

We could add a CI workflow that updates the pinned version and auto-merges if our checks succeed (we have enough test coverage to do that now).

Another idea would be for the installer to install luarocks into a temp directory (or to use the system installation if detected)
and for rocks.nvim to add [luarocks](https://luarocks.org/modules/hisham/luarocks) to its dependencies.
This needs to be investigated and tested.

mrcjkb:
Closes #336

### DIFF
--- a/installer.lua
+++ b/installer.lua
@@ -349,15 +349,18 @@ local function install()
         if line == "< OK >" then
             local install_path = input_fields.install_path.content
             local setup_luarocks = input_fields.setup_luarocks.content == "true"
+            local temp_luarocks_path =
+                ---@diagnostic disable-next-line: param-type-mismatch
+                vim.fs.joinpath(vim.fn.stdpath("run"), ("luarocks-%X"):format(math.random(256 ^ 7)))
 
             local luarocks_binary = "luarocks"
 
             if setup_luarocks then
-                local success = set_up_luarocks(install_path)
+                local success = set_up_luarocks(temp_luarocks_path)
                 if not success then
                     return
                 end
-                luarocks_binary = vim.fs.joinpath(install_path, "bin", "luarocks")
+                luarocks_binary = vim.fs.joinpath(temp_luarocks_path, "bin", "luarocks")
             elseif vim.fn.executable(luarocks_binary) ~= 1 then
                 vim.notify(
                     luarocks_binary
@@ -389,7 +392,6 @@ local function install()
 
             acquire_buffer_lock(buffer, function()
                 local install_path_rel = install_path:gsub(vim.env.HOME, "")
-                local luarocks_binary_rel = luarocks_binary:gsub(vim.env.HOME, "")
 
                 vim.api.nvim_buf_set_lines(buffer, 0, -1, true, {
                     "INSTALLATION COMPLETE",
@@ -399,7 +401,6 @@ local function install()
                     ">lua",
                     " local rocks_config = {",
                     '     rocks_path = vim.env.HOME .. "' .. install_path_rel .. '",',
-                    '     luarocks_binary = vim.env.HOME .. "' .. luarocks_binary_rel .. '",',
                     " }",
                     " ",
                     " vim.g.rocks_nvim = rocks_config",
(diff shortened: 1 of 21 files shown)


mrcjkb:
The bootstrapping guide still hard codes a luarocks install. Not sure how to solve that.

installer.lua
```diff
@@ -398,8 +397,6 @@ local function install()
                     "The code has already been copied to your clipboard:",
                     ">lua",
                     " local rocks_config = {",
-                    '     rocks_path = vim.env.HOME .. "' .. install_path_rel .. '",',
```
NTBBloodbath, line 401: question: was this removal by mistake? This line is still used below 👀

vhyrro:
Code written thus far looks good to me apart from the removal NTB mentioned.

In my opinion, we should rewrite a large portion of the installation script to perform a temporary installation into `/tmp` for UNIX systems. The bootstrapped luarocks should only serve as a temporary vessel for the one actually installed into the tree. To make the experience smooth, the temporary luarocks should be passed the same `--tree` as the location where the new luarocks will be installed; the freshly installed luarocks should also use this same tree therefore as well.

I think we should try to tackle the problem in one big swoop here, it's quite likely we'll need to mark these changes as breaking and bump to 3.0? I can imagine that the `vim.g.rocks` might need some slight changes.

installer.lua
mrcjkb: Yep, that was unintended

mrcjkb:
I've updated the installer to install to a temp directory.

> I think we should try to tackle the problem in one big swoop here, it's quite likely we'll need to mark these changes as breaking and bump to 3.0? I can imagine that the vim.g.rocks might need some slight changes.

I don't think actually breaks any existing installations. And it might be nice to still be able to override the `luarocks_binary` in the config (e.g. when using nix).
But removing the luarocks install from existing installations would most likely be a manual process, so a major version bump might be in order. This would also give us an opportunity to remove the deprecated API.

mrcjkb:
I can't figure out why the nix build of rocks.nvim fails with

```
┃        > Missing dependencies for rocks.nvim scm-1:
┃        >    luarocks (not installed)
┃        >
┃        > rocks.nvim scm-1 depends on lua >= 5.1 (5.1-1 provided by VM: success)
┃        > rocks.nvim scm-1 depends on luarocks (not installed)
┃        > Warning: Failed searching manifest: Failed downloading https://luarocks.org/manifest-5.1 - no downloader tool available, please in…
┃        >
┃        > Error: Could not satisfy dependency luarocks: No results matching query were found for Lua 5.1.
┃        For full logs, run 'nix log /nix/store/arbpyscrxmvx2xyj9hhv77rmwrzf28b4-luajit2.1-rocks.nvim-scm-1.drv'.
```

@teto do you have an idea? Could this be an issue with `build-luarocks-package.nix`?

mrcjkb:
https://github.com/NixOS/nixpkgs/issues/316009

mrcjkb:
This one is technically ready, but I would like to wait for a luarocks release that has a fix for https://github.com/luarocks/luarocks/pull/1661

mrcjkb:
...and of course it's broken on Windows :angry:

mrcjkb:
So... on Windows, luarocks installs a  `rocks_path/bin/luarocks.bat` binary when installing the `luarocks` package.
This doesn't seem to work with `vim.system`, as it doesn't accept unix-like CLI args.

I'm done with this for Windows :angry:. We have a few options:

- Request that the luarocks rock install a proper `luarocks` executable, equivalent to what you get on Windows when installing luarocks regularly.
- Tell Windows users to use a system installation and keep it up-to-date
- Implement a special case for the installer (which doesn't have proper Windows support yet), where we don't install luarocks into a temp directory on Windows.

teto:
sry I completely missed your ping here. The nix issue will be fixed with https://github.com/NixOS/nixpkgs/pull/316221 hopefully

(dropping this here https://github.com/oploadk/localua as it looked of interest).

vhyrro:
Code LGTM, although I've yet to actually try this on my machine, so that'll be the next step :)

vhyrro (approved):
Did a crude manual test but it seems to function nicely :)

PR #353, opened 2024-05-26
fix(operations): always force reinstalls if rocks are already installed

### LINKED ISSUE #351: Investigate `Rocks update` potentially not updating `scm`/`dev` plugins
I guess `:Rocks update` didn't update the dev plugin, which may be another bug I need to investigate.

_Originally posted by @mrcjkb in https://github.com/nvim-neorocks/rocks-treesitter.nvim/issues/12#issuecomment-2132173321_

mrcjkb:
Fixes #351 

Note: This doesn't actually work with luarocks 3.11.0, see https://github.com/luarocks/luarocks/issues/1657

### DIFF
--- a/lua/rocks/operations/helpers.lua
+++ b/lua/rocks/operations/helpers.lua
@@ -61,6 +61,7 @@ helpers.install = function(rock_spec, progress_handle)
             table.insert(install_cmd, version)
         end
     end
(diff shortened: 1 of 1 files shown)


vhyrro (approved):
Should we wait for a new release of luarocks to drop, or are we merging this as-is?

PR #389, opened 2024-06-17
fix(api): include `dev` rocks in `query_luarocks_rocks`

mrcjkb:


mrcjkb:
This one is quite trivial. Merging.

mrcjkb:
Integration tests are failing because lua.org is down.

PR #392, opened 2024-06-18
feat(api): `RockSpecModifier` preload hooks

mrcjkb:
This refines the preload hook API to support different types of hooks.

For now, the following are supported:

- `RockSpecModifier`s, which allow extensions to modify the hooks that rocks.nvim sees.
  - rocks-lazy.nvim will be able to use this to let rocks.nvim know that a plugin is `opt`, without users having to set the `opt` field.
  - This could potentially also be useful for the rocks-nix.nvim module I'm planning.
- `Action`s, which is what we have already implemented in rocks-config.nvim, only that it can now take the (potentially modified) `user_rocks: RockSpec[]` as a parameter, so that rocks-config.nvim doesn't have to read it from the rocks.toml again.

The modifier does not do anything with the `MutRocksTomlRef` (the one we use to write to rocks.toml on install, update, ...). I think it's good that it doesn't.

### DIFF
--- a/doc/rocks.txt
+++ b/doc/rocks.txt
@@ -260,15 +260,41 @@ rocks.nvim API hooks                                           *rocks.api.hooks*
 Hooks that rocks.nvim modules can inject behaviour into.
 Intended for use by modules that extend this plugin.
 
+ Preload hooks                                                *rocks.hooks.preload*
 
-
- |preload|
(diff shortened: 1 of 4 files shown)


mrcjkb:
TODO:

- [ ] Come up with some automated test cases.

PR #409, opened 2024-06-24
Update README.md

craigmac:


mrcjkb:
Thanks 🙏

https://github.com/nvim-neorocks/rocks.nvim/issues/386 might make this redundant, so I'd rather focus on that.

mrcjkb:
@craigmac https://github.com/nvim-neorocks/rocks.nvim/issues/386 is proving more complicated than I though.

The luarocks binary has a `#!/usr/bin/env lua` shebang, so setting `variables.LUA` in the config won\t solve this.

PR #411, opened 2024-06-24
feat: configure luarocks to use Neovim as the Lua interpreter

mrcjkb:


mrcjkb:
Problem: The luarocks installation bundled with luarocks has a `#!/usr/bin/env lua` shebang.

So bundling `nlua` won't prevent that from throwing an error if `lua` is missing :disappointed:

PR #423, opened 2024-06-28
remove myself from codeowners

teto:
I am likely to get more involved with the projects but I dont like to be autopinged. Ping me if you need outside outlook else I trust you

I also removed @NTBBloodbath as discussed on channel: give the man some rest :) I can drop it i state of mind has changed.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,5 +11,5 @@ spec/ @NTBBloodbath @vhyrro @mrcjkb
 plugin/ @NTBBloodbath @vhyrro @mrcjkb
 
 # Nix
-nix/ @mrcjkb @teto
-flake.* @mrcjkb @teto
+nix/ @mrcjkb
+flake.* @mrcjkb


vhyrro:
Ah that's fair enough - the amount of pings is quite large indeed :p

@NTBBloodbath what's your take? Let me know what the course of action should be: do we keep you just for documentation (.md files), for everything (including code) or for nothing?

teto:
I find the github action bot pretty noisy. Can it be disabled for contributors of the repo already ? they should know the rules already.

NTBBloodbath:
> Ah that's fair enough - the amount of pings is quite large indeed :p
> 
> @NTBBloodbath what's your take? Let me know what the course of action should be: do we keep you just for documentation (.md files), for everything (including code) or for nothing?

I've already been recovering from burnout and the notifications don't bother me, so it's okay to keep me in the codeowners for everything. Thank you both very much for caring and asking about this :)

PR #425, opened 2024-06-28
doc: add default for generate_help_pages

teto:
tried to harmoinze the display of defaults as well

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -127,6 +127,7 @@
               # to be generated
               gcc
               tree-sitter
+              docgen
             ])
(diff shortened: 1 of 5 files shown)


teto:
ha I see it's generated ^^'

flake.nix
```diff
@@ -127,6 +127,7 @@
               # to be generated
               gcc
               tree-sitter
+              docgen
```
teto, line 130: I prefer to run docgen from within the shell then `nix run .#docgen` which is slower (it copies the fodler to the store + nix eval time).

PR #430, opened 2024-06-28
fix(health): do not produce duplicated messages when there are no errors in the `rocks.toml` file

### LINKED ISSUE #429: checkhealth has redundant output
with rocks 2.31 `checkhealth rocks`  gives:
```
Checking rocks.toml ~
- OK No errors found in rocks.toml.
- OK No errors found in rocks.toml.
- OK No errors found in rocks.toml.
- OK No errors found in rocks.toml.
- OK No errors found in rocks.toml.
- OK No errors found in rocks.toml.
```
my rocks.toml:
```
# This is your rocks.nvim plugins declaration file.
# Here is a small yet pretty detailed example on how to use it:
#
# [plugins]
# nvim-treesitter = "semver_version"  # e.g. "1.0.0"

# List of non-Neovim rocks.
# This includes things like `toml` or other lua packages.
[rocks]

# List of Neovim plugins to install alongside their versions.
# If the plugin name contains a dot then you must add quotes to the key name!
[plugins]
# "rocks.nvim" = "2.19.0" # rocks.nvim can also manage itself :D
cmp-rg = "1.3.9"
"oil.nvim" = "2.10.0"
# -- :DiffviewOpen
"diffview.nvim" = "scm"
"haskell-tools.nvim" = "3.1.10"
"none-ls.nvim" = "scm"
"bufferline.nvim" = "4.6.1"

[config]
plugins_dir = "plugins/"
```

I think I would like maybe a more in detail command, like `:Rocks debug` or `:Rocks check` that displays the path towards LUAROCKS_CONFIG, luarocks --version and so on. Maybe the rocks.nvim extensions too if there is a way to identify those.

NTBBloodbath:
Fixes #429.

I think the way the fix was done is simple yet works well, but if there are any suggestions it will be more than welcome :)

### DIFF
--- a/lua/rocks/health.lua
+++ b/lua/rocks/health.lua
@@ -133,8 +133,10 @@ end
 
 local function check_rocks_toml()
     start("Checking rocks.toml")
+    local found_err = false
     local success, user_rocks_or_err = xpcall(require("rocks.config.internal").get_user_rocks, function(err)
         error(err)
+        found_err = true
     end)
     if not success then
         return
@@ -142,7 +144,10 @@ local function check_rocks_toml()
     for rock_name, _ in pairs(user_rocks_or_err) do
         if rock_name:lower() ~= rock_name then
             error(("Plugin name is not lowercase: %s"):format(rock_name))
+            found_err = true
         end
+    end
+    if not found_err then
         ok("No errors found in rocks.toml.")
     end
 end


mrcjkb (approved):

teto:
tested locally and it fixed the issue ty !

PR #432, opened 2024-06-28
feat: make errors more visible with vim.notify

teto:
the spinner is too discrete and there is a green tick once it finishes so at first I thought the install had succeeded but paying closer attention, the message up/down of it said install had failed

I haven't tested yet. Just putting it out there to see what people think


mrcjkb (commented):
I suppose we'd also want this for `:Rocks update`?

vhyrro:
Wait, why is there a green tick at the end, if there were errors during the process? I feel like that should be the meat of this problem. I personally feel like a notify and a message in fidget.nvim is slightly too much :p

PR #435, opened 2024-06-28
doc: introduce rocks.troubleshooting + UX tweaks

teto:
Looking at https://github.com/nvim-neorocks/rocks.nvim/issues/424, I think we can better guide the user in case of error, instead of displaying `Could not find diffview.nvim. Search for 'dev' version? y/n: `, we could show a vim.ui.select:
1. Install the "dev" version (when we know it exists)
2. Install via rocks-git (and install rocks-git if not available)
< mention rocks.troubleshootingi or an url towards how to add plugins to luarocks / link to https://github.com/nvim-neorocks/rocks.nvim/issues/111 >

### DIFF
--- a/doc/rocks.txt
+++ b/doc/rocks.txt
@@ -71,6 +71,12 @@ RocksOpts                                                            *RocksOpts*
         {luarocks_config?}                (string)   Path to the luarocks config. If not set, rocks.nvim will create one in `rocks_path`. Warning: You should include the settings in the default luarocks-config.lua before overriding this.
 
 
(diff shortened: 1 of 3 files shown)


mrcjkb:
> Install via rocks-git (and install rocks-git if not available)

I think it's fine just to mention the possibility of using rocks-git in the docs.
rocks.nvim shouldn't be "aware" of any external modules and there's no way to infer the repo owner.

teto:
fine

PR #436, opened 2024-06-29
fix(api): use user rocks with preload hook modifications applied

mrcjkb:


mrcjkb:
Merging, as this is trivial.

@NTBBloodbath this should fix the issue with having to set `opt=true` for rocks-lazy.nvim to work.
No idea why it is working for me in my Nix derivation :shrug:

PR #450, opened 2024-07-02
feat(log): trace luarocks_config path

teto:
To help debug/diagnose if config is correctly set. It can be deceiving in case vim.g.rocks_nvim gets changed later.

Which is why at first I thought of passing the luarocks config via the command line instead of the environment but luarocks still doesn't support it: https://github.com/luarocks/luarocks/issues/1086

This looks like a good compromise.

### DIFF
--- a/plugin/rocks.lua
+++ b/plugin/rocks.lua
@@ -24,6 +24,7 @@ log.trace("loading rocks.adapter")
 local adapter = require("rocks.adapter")
 log.trace("loading rocks config")
 local config = require("rocks.config.internal")
+log.debug("Using luarocks config " .. config.luarocks_config)
 
 -- Initialize the luarocks loader
 if config.enable_luarocks_loader then


plugin/rocks.lua
```diff
@@ -18,6 +18,7 @@ end
 
 local log = require("rocks.log")
 log.info(vim.uv.os_uname())
+log.info("Using luarocks config " .. (vim.g.rocks_nvim.luarocks_config or "unset"))
```
mrcjkb, line 21: issue: we should get the luarocks config from the `config` module (see L27)

mrcjkb (changes requested):

plugin/rocks.lua
teto: ha yes ty. I noticed that after the initial implementation and forgot to correct it. The CI error looks unrelated

mrcjkb (approved):

PR #451, opened 2024-07-02
fix(prune): need to prune twice for rocks-git to remove plugin

mrcjkb:
Fixes https://github.com/nvim-neorocks/rocks-git.nvim/issues/34


mrcjkb:
Merging, as this is trivial

Issue #443, opened 2024-07-01
[Feature] `Rocks install` API for `install_args`


mrcjkb:
> For instance, allowing :Rocks install something `opt=true` `ft=markdown` and feeding that directly into the toml.

We already do that[^1] :wink: 

[^1]: As long as the extensions know how to parse args that rocks.nvim doesn't know about.
       So far, only rocks.nvim and rocks-git.nvim know how to parse args.

What could be tricky is coming up with a good API for args that are lists:

```vim
:Rocks install {rock} opt=true install_args=["GIT2_DIR=/some/path","foo","bar"]
```

or

```vim
:Rocks install {rock} opt=true install_args=[GIT2_DIR=/some/path,foo,bar]
```

or, assuming any args that rocks.nvim can't parse are `install_args`:

```vim
:Rocks install {rock} opt=true GIT2_DIR=/some/path
```

The last one is more concise, but also more brittle: If someone passes in `ft=markdown`, it would be added to the `install_args` as long as rocks-lazy.nvim doesn't register an arg parser.

> should we instead support an [env] section in the toml, which would modify vim.env?

Luarocks won't pick up `GIT2_DIR` from an environment variable. It has to be passed as an arg to `install`.

Since this is more of an edge case, I think we don't need to prioritise passing `install_args` via `:Rocks install` initially.

_Originally posted by @mrcjkb in https://github.com/nvim-neorocks/rocks.nvim/issues/439#issuecomment-2200585803_

mrcjkb:
Since this ought to be an edge case, we may not want to support this, as it adds quite a bit of complexity.

PR #442, opened 2024-07-01
feat(rocks.toml): support specifying extra luarocks `install_args`

mrcjkb:
Closes #439

PR #452, opened 2024-07-02
feat: support extending the default luarocks config

mrcjkb:
Stacked on #442 

This adds a solution for #443 by making it easier to defend the default luarocks config.

You can add extra directories to search for binaries via

```lua
vim.g.rocks_nvim = {
  -- ...
  ---@type string | table file path or config table to extend the default with
  luarocks_config = {
    external_deps_dirs = { "/some/path" },
  },
}
```

See https://github.com/luarocks/luarocks/wiki/Config-file-format for details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-result
+result*
 https___*
 .pre-commit-config.yaml
 .direnv
--- a/plugin/rocks.lua
+++ b/plugin/rocks.lua
@@ -28,7 +28,7 @@ log.debug("Using luarocks config " .. config.luarocks_config)
 
 -- Initialize the luarocks loader
 if config.enable_luarocks_loader then
-    require("rocks.loader").enable()
+    nio.run(require("rocks.loader").enable)
 end
 
 -- Set up the Rocks user command
--- a/plugin/rocks.lua
+++ b/plugin/rocks.lua
@@ -24,11 +24,12 @@ log.trace("loading rocks.adapter")
 local adapter = require("rocks.adapter")
 log.trace("loading rocks config")
 local config = require("rocks.config.internal")
-log.debug("Using luarocks config " .. config.luarocks_config)
 
 -- Initialize the luarocks loader
 if config.enable_luarocks_loader then
-    nio.run(require("rocks.loader").enable)
+    nio.run(function()
+        require("rocks.loader").enable()
+    end)
 end
 
 -- Set up the Rocks user command
--- a/plugin/rocks.lua
+++ b/plugin/rocks.lua
@@ -27,9 +27,7 @@ local config = require("rocks.config.internal")
 
 -- Initialize the luarocks loader
 if config.enable_luarocks_loader then
-    nio.run(function()
-        require("rocks.loader").enable()
-    end)
+    require("rocks.loader").enable()
 end
 
 -- Set up the Rocks user command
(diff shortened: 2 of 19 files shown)


.gitignore
```diff
@@ -1,4 +1,6 @@
 result
+result-dev
+result-lib
```
teto, line 3: `result*` ?

mrcjkb:
> unrelated to this PR but I noticed some redundancy between lua/rocks/config/init.lua and lua/rocks/config/internal.lua RocksOpts and RocksConfig : aren't these the same ? can we do something about the doc redundancy ?

`config/init` and `RocksOpts` are exposed to the user. Each field of `RocksOpts` is optional, allowing users to omit fields without getting lua-language-server warning diagnostics.

`config/internal` and `RocksConfig` are not exposed to the user. None of the fields of `RocksConfig` are optional, giving us better type safety.

mrcjkb:
I've made writing/loading the luarocks config async + on demand. That ought to minimise the impact on startup time.
Also, if we need to update the config at runtime (maybe for parallel installs?), this is the better approach.

teto:
I dont know what to think here. The `luarocks_config` as a string/table makes things more complex to understand.
 
If it's a string, then we just load the file without merging.
If it's a table, then we merge and write the file (but can't decide the location anymore).

Also it's user-facing so if it ends being a bad idea, it's hard to rewind.

One of the problems is that our luarocks config depends on  `vim.g.rocks_nvim.rocks_path`, that seems like the only reason for merging the configs. We could mention that the first rock_tree is where to install stuff ? but that might be too complex for users.

If I take the nix scenario, I can generate the variables/rockspecs and pass it as `vim.g.rocks_nvim.luarocks_config` table. Right now, I had to copy the defaults from rocks.nvim into my generated luarocks config file so in that regards, the PR improves things.

mrcjkb:
> If it's a string, then we just load the file without merging.
If it's a table, then we merge and write the file (but can't decide the location anymore).

The string is to prevent this from being a breaking change.
If we decide we don't want to support it anymore, we can start off with a `vim.deprecate` and then remove support for it being a string in version 3.

mrcjkb:
How on earth did the integration tests get flaky again after a rebase?

plugin/rocks.lua
```diff
@@ -24,11 +24,12 @@ log.trace("loading rocks.adapter")
...
 log.trace("loading rocks config")
 local config = require("rocks.config.internal")
-log.debug("Using luarocks config " .. config.luarocks_config)
 
 -- Initialize the luarocks loader
 if config.enable_luarocks_loader then
-    require("rocks.loader").enable()
+    nio.run(function()
+        require("rocks.loader").enable()
+    end)
```
vhyrro, line 32: question: is there ever a situation where this could break certain configurations that try to make use of the loader's features directly on startup? If there's a possibility of a race condition where the loader is enabled after it is used then we might want to reconsider making this behaviour asynchronous.

mrcjkb: I have been wondering about this too.

There is some strange behaviour I observed in the integration test: 

```bash
nvim -u .github/resources/init-integration.lua -c +wq
```

now exits with a non-zero exit code. But it does what it's supposed to do.
My guess is that it reaches `+q` before it finishes writing the luarocks config. 

I think it's safer not to make it async.

mrcjkb: Done.

mrcjkb:
Merging, as tests are succeeding and all issues have been addressed.

PR #456, opened 2024-07-03
docs(readme): update

teto:
see commit messages but to sum up:
- moved down ``Cargo-like rocks.toml file for declaring all your plugins` since I find this a weak feature, less powerful for sure than what was coming next aka "automatic dependency management"
- Added a section to mention the competition

I dont like that we mention twice the ecosystem about all the extensions but not sure how to do better so I left it untouched.
I do think that section should appear in `:h rocks.nvim` (currently is the barebone `A luarocks plugin manager for Neovim.` xD) because users are bound to want to install plugins from git so at least rocks-git.nvim should appear in help IMO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,12 @@
 
 ## :star2: Features
 
+- Automatic dependency and build script management.
+- True [semantic versioning](https://semver.org/)!
 - `Cargo`-like `rocks.toml` file for declaring all your plugins.
 - Name-based installation
   (` "nvim-neorg/neorg" ` becomes `:Rocks install neorg` instead).
-- Automatic dependency and build script management.
 - Supports [multiple versions of the same dependency](https://github.com/luarocks/luarocks/wiki/Using-LuaRocks#multiple-versions-using-the-luarocks-package-loader).
-- True semver versioning!
 - Minimal, non-intrusive UI.
 - Async execution.
 - Extensible, with a Lua API.
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@
     for lazy-loading.
   - [`rocks-treesitter.nvim`](https://github.com/nvim-neorocks/rocks-treesitter.nvim)
     for automatic tree-sitter parser management.
-  - And more...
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -498,6 +500,13 @@ which contains the `luarocks` stderr output, among other logs.
 - [luarocks.nvim](https://github.com/vhyrro/luarocks.nvim):
   Adds basic support for installing lua rocks to [lazy.nvim](https://github.com/folke/lazy.nvim)
 
+
+## Other neovim plugin managers
```
mrcjkb, line 504:
```suggestion
## :link: Other neovim plugin managers
```

README.md
```diff
@@ -498,6 +500,13 @@ which contains the `luarocks` stderr output, among other logs.
...
+
+- [lazy.nvim](https://github.com/folke/lazy.nvim): started luarocks support in
+  version [11.X](https://lazy.folke.io/news#11x)
+- [vim-plug](https://github.com/junegunn/vim-plug)
+
```
mrcjkb, line 509: note: It feels a bit strange to me mentioning only lazy.nvim and vim-plug here.
We should either move the lazy.nvim mention to the above section.
Or we should mention additional plugin mangers.
Some that come to mind:

- [pckr](https://github.com/lewis6991/pckr.nvim)
- [paq-nvim](https://github.com/savq/paq-nvim)
- [mini.deps](https://github.com/echasnovski/mini.deps)

mrcjkb (changes requested):

README.md
teto: absolutely. Do you think I should merge this section with the previous ?

teto: I think I like the separation though because it makes it clear those are "competition"

teto: similar type of rocks.nvim. Could be a subsection of the previous one as well.

mrcjkb: I think two sections are fine.

mrcjkb (approved):

Issue #439, opened 2024-07-01
[Feature] Ability to pass extra luarocks installation args


SuperBo:
Hi, when installing https://github.com/SuperBo/fugit2.nvim with Rocks `:Rocks install fugit2.nvim`. I met these errors.

```
Error: Could not find library file for GIT2
  No file /git2/lib/libgit2.dylib in /usr/local/lib
  No file /git2/lib/libgit2.dylib in /usr/local
  No file libgit2.tbd in /usr/local/lib
  No file libgit2.tbd in /usr/local
  No file libgit2.a in /usr/local/lib
  No file libgit2.a in /usr/local
  No file libgit2.dylib in /usr/local/lib
  No file libgit2.dylib in /usr/local
  No file matching libgit2.so.* in /usr/local/lib
  No file matching libgit2.so.* in /usr/local
  No file /git2/lib/libgit2.dylib in /usr/lib
  No file /git2/lib/libgit2.dylib in /usr
  No file libgit2.tbd in /usr/lib
  No file libgit2.tbd in /usr
  No file libgit2.a in /usr/lib
  No file libgit2.a in /usr
  No file libgit2.dylib in /usr/lib
  No file libgit2.dylib in /usr
  No file matching libgit2.so.* in /usr/lib
  No file matching libgit2.so.* in /usr
  No file /git2/lib/libgit2.dylib in /lib
  No file /git2/lib/libgit2.dylib in /
  No file libgit2.tbd in /lib
  No file libgit2.tbd in /
  No file libgit2.a in /lib
  No file libgit2.a in /
  No file libgit2.dylib in /lib
  No file libgit2.dylib in /
  No file matching libgit2.so.* in /lib
  No file matching libgit2.so.* in /
  No file /git2/lib/libgit2.dylib in /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib
  No file /git2/lib/libgit2.dylib in /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr
  No file libgit2.tbd in /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib
  No file libgit2.tbd in /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr
  No file libgit2.a in /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib
[...]

PR #459, opened 2024-07-04
feat: prompt to update plugins with breaking changes + options to skip prompts with `!`

mrcjkb:
Closes #191 

Stacked on #439 (already reviewed by @teto) to avoid merge conflicts.

### DIFF
--- a/lua/rocks/commands.lua
+++ b/lua/rocks/commands.lua
@@ -140,7 +140,7 @@ local rocks_command_tbl = {
                 return
             end
             require("rocks.operations").add(args, nil, {
-                auto_search_dev = opts.bang,
+                skip_prompts = opts.bang,
             })
         end,
         complete = function(query)
--- a/lua/rocks/commands.lua
+++ b/lua/rocks/commands.lua
@@ -124,7 +124,7 @@ local rocks_command_tbl = {
     update = {
         impl = function(_, opts)
             require("rocks.operations").update(nil, {
-                skip_prompt = opts.bang,
+                skip_prompts = opts.bang,
             })
         end,
     },
--- a/lua/rocks/commands.lua
+++ b/lua/rocks/commands.lua
@@ -16,12 +16,13 @@
 ---                                       - pin={true|false}
 ---                                         Rocks that have been installed with 'pim=true'
 ---                                         will be ignored by ':Rocks update'.
+---                                     Use 'Rocks! install ...' to skip prompts
 --- prune {rock}                        Uninstall {rock} and its stale dependencies,
 ---                                     and remove it from rocks.toml.
 --- sync                                Synchronize installed rocks with rocks.toml.
 ---                                     It may take more than one sync to prune all rocks that can be pruned.
 --- update                              Search for updated rocks and install them.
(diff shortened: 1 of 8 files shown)


lua/rocks/operations/helpers.lua
```diff
@@ -176,4 +176,74 @@ helpers.is_installed = nio.create(function(rock_name)
...
 
+---@class RockUpdate
+---@field name rock_name
+---@field version vim.Version
+---@field target_version vim.Version
+---@field pretty string
+
+---@param outdated_rocks table<rock_name, OutdatedRock>
+---@return table<rock_name, RockUpdate[]>
+function helpers.get_breaking_changes(outdated_rocks)
```
mrcjkb, line 187: TODO: 

- [ ] filter out rocks that have been pinned

mrcjkb:
Maybe we should also check for major version bumps when installing a rock that is already installed

lua/rocks/commands.lua
```diff
@@ -119,8 +122,10 @@ end
 ---@type { [string]: RocksCmd }
 local rocks_command_tbl = {
     update = {
-        impl = function(_)
-            require("rocks.operations").update()
+        impl = function(_, opts)
+            require("rocks.operations").update(nil, {
+                skip_prompt = opts.bang,
```
teto, line 127: bikeshedding: skip_prompt -> force_major_updates ? or skip_semver_prompt.

mrcjkb: That's a typo. I renamed it to `skip_prompts` (plural), because there could be more than one prompt, as is the case with `install`.

mrcjkb, line 127:
```suggestion
                skip_prompts = opts.bang,
```

lua/rocks/commands.lua
```diff
@@ -129,12 +134,14 @@ local rocks_command_tbl = {
...
     install = {
-        impl = function(args)
+        impl = function(args, opts)
             if #args == 0 then
                 vim.notify("Rocks install: Called without required package argument.", vim.log.levels.ERROR)
                 return
             end
-            require("rocks.operations").add(args)
+            require("rocks.operations").add(args, nil, {
+                auto_search_dev = opts.bang,
```
mrcjkb, line 143:
```suggestion
                skip_prompts = opts.bang,
```

lua/rocks/operations/helpers.lua
```diff
@@ -176,4 +176,96 @@ helpers.is_installed = nio.create(function(rock_name)
...
+    if type(version) == "table" and type(target_version) == "table" and target_version.major > version.major then
+        return setmetatable({
+            name = rock.name,
+            version = version,
+            target_version = target_version,
+        }, {
+            __tostring = function()
+                return ("%s %s -> %s"):format(rock.name, tostring(version), tostring(target_version))
+            end,
+        })
```
vhyrro, line 193: praise: love this use of metatables, wouldn't have thought of that

vhyrro (approved):
Couldn't find anything to complain about. Code looks good to me, I like this change a lot :)

PR #462, opened 2024-07-06
refactor(operations): split into smaller modules

mrcjkb:


vhyrro (approved):
Makes code much more readable indeed, LGTM!

PR #485, opened 2024-07-27
feat: update remote plugins after install/update

mrcjkb:


vhyrro (approved):

teto:
I am curious what was the request asking for it . Note that justin seems to be actively trying to remove remote plugins so maybe this could have been a RocksPostSync hook instead.

mrcjkb:
> I am curious what was the request asking for it . Note that justin seems to be actively trying to remove remote plugins so maybe this could have been a RocksPostSync hook instead.

There wasn't a request. I just realised that remote plugins are still a thing and they typically tell people to put `build = ':UpdateRemotePlugins' ` in their install instructions.
Since the command only needs to be run once after install/update/sync, it doesn't have a significant impact on performance and I see no harm in adding it.

teto:
yes it's a good idea but I saw you were going to release rocks.nvim and I thought I would suggest a more generic event (thought not sure how one would use that) that could survive a breaking change.

mrcjkb:
I think user events could be a good idea.
rocks.nvim extensions could hook into them too.
We could also have events for `RocksFooCachePopulated` or `RocksFooCacheUpdated`, which might be useful for rocks-edit.nvim.

Let's create an issue for it and collect ideas.

mrcjkb:
https://github.com/neovim/neovim/issues/27949

PR #492, opened 2024-07-29
feat(experimental): a way to recognise plugins managed by extensions as dependencies

### LINKED ISSUE #365: [Feature] plugins installed with rocks-git, rocks-dev, ... should be recognized as dependencies
This could potentially be achieved by installing a dummy package to spoof luarocks into thinking the rock is installed as a luarocks package.

mrcjkb:
Closes #365 

This provides a function that extensions can invoke to install or remove a luarocks stub, in order to trick luarocks into thinking a luarocks package is installed.

That way, plugins managed by extensions like rocks-git or rocks-dev can be recognised by luarocks as being installed, if other luarocks packages depends on them.

TODO:

- [x] Use the API in rocks-git and test it manually.
- [x] Use the API in rocks-dev and test it manually.

Requires the following experimental feature to be enabled:

```lua
vim.g.rocks_nvim = {
  -- ...
  experimental_features = {
    "ext_module_dependency_stubs",
  },
}
```

and up-to-date rocks.nvim extensions.

### DIFF
--- a/doc/rocks.txt
+++ b/doc/rocks.txt
@@ -14,6 +14,7 @@ rocks.nvim ··································�
  ························································· |rocks-toml.luarocks|
 rocks.nvim commands ··········································· |rocks-commands|
 rocks.nvim configuration ········································ |rocks-config|
+experimental features ····································· |rocks.experimental|
 rocks.nvim |User| |event|s ·································· |rocks.user-event|
 Lua API for rocks.nvim extensions ·································· |rocks-api|
 rocks.nvim API hooks ········································· |rocks-api-hooks|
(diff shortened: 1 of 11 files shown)


mrcjkb:
Done and tested.

We could later add the ability to add dependencies to the stub template; see https://github.com/nvim-neorocks/rocks.nvim/issues/388

mrcjkb:
come to think of it, we should probably put this behind a config flag and make it an experimental feature, with an announcement calling for testers.

doc/rocks.txt
```diff
@@ -138,31 +139,41 @@ RocksOpts                                                            *RocksOpts*
...
+        {experimental_features?}          (rocks.ExperimentalFeature[])
+                                                                          List of experimental features to enable.
+                                                                          See |rocks.experimental|.
+
+
+==============================================================================
+experimental features                                       *rocks.experimental*
+
+WARNING: Experimental features are not considered part of the official API.
+They may change or be removed without a major SemVer version bump.
```
mrcjkb, line 177: For experimental features to be documented, we need https://github.com/mrcjkb/vimcats/issues/17

mrcjkb:
Self-merging, as this is now experimental and disabled by default.

PR #494, opened 2024-08-02
chore: some rocks-binaries improvements and fixes

### LINKED ISSUE #342: doesn't work on systems without libgcc_s.so.1
On Linux distributions without libgcc, such as [Chimera Linux](https://chimera-linux.org/) which uses libunwind instead, nvim spits the following error when opening nvim.
```
Error executing Lua callback: ...<user>/.local/share/nvim/rocks/share/lua/5.1/nio/tasks.lua:100: Async task failed without callback: The coroutine failed with this message:
error loading module 'toml_edit' from file '/home/<user>/.local/share/nvim/rocks/lib/lua/5.1/toml_edit.so':
        Error loading shared library libgcc_s.so.1: No such file or directory (needed by /home/<user>/.local/share/nvim/rocks/lib/lua/5.1/toml_edit.so)
stack traceback:
        [C]: at 0x7457690eac80
        [C]: in function 'require'
        ...share/nvim/rocks/share/lua/5.1/rocks/config/internal.lua:55: in function 'get_rocks_toml'
        ...share/nvim/rocks/share/lua/5.1/rocks/config/internal.lua:75: in function 'get_user_rocks'
        .../luarocks/rocks-5.1/rocks.nvim/2.28.0-1/plugin/rocks.lua:57: in main chunk
```                                                                                                                                 
It seems might be an issue with the toml-edit dependency, but I am unsure. Is there anything that can be done? Or is the issue with luarocks linking libgcc?

### LINKED ISSUE #441: [Feature] Allow overriding or disabling the `rocks-binaries` server
This could be done 

- Per plugin  with a `servers=["<server>"]` or `servers=false` option.
- Or globally, via a `[servers]` option, which overrides rocks-binaries if it is set (disabling it if it's empty).

Ditto for `dev-servers`.

### LINKED ISSUE #487: rocks-binaries should not be used on musl systems (e.g. Alpine)
On systems using musl, binaries from rocks-binaries will not work by default, and there's no way to force luarocks to compile from source at the current moment (e.g. a config option). As an example, if you set up a barebones rocks.nvim installation on a musl system and then run `:Rocks install tree-sitter-bash`, you will end up with a binary linked against glibc that will therefor fail to load:

```
vi@virtuoso ~ % cd .local/share/nvim/rocks/lib/lua/5.1/parser 
vi@virtuoso parser % lddtree bash.so 
bash.so => ./bash.so (interpreter => none)
    libc.so.6 => not found
```

Or for another example, if you reinstall or update toml-edit through rocks.nvim after bootstrapping, then it will attempt to use the binary linked against glibc and throw this error on startup:
```
Error detected while processing /home/vi/.local/share/nvim/rocks/lib/luarocks/rocks-5.1/rocks.nvim/2.36.1-1/plugin/rocks.lua:
E5113: Error while calling lua chunk: ...local/share/nvim/rocks/rocks_rtp/lua/luarocks/loader.lua:104: error loading module 'toml_edit' from file '/home/vi/.local/share/nvim/rocks/lib/lua/5.1/toml_edit.so':
        Error loading shared library ld-linux-x86-64.so.2: No such file or directory (needed by /home/vi/.local/share/nvim/rocks/lib/lua/5.1/toml_edit.so)
stack traceback:
        [C]: in function 'a_loader'
        ...local/share/nvim/rocks/rocks_rtp/lua/luarocks/loader.lua:104: in function <...local/share/nvim/rocks/rocks_rtp/lua/luarocks/loader.lua:101>
        [C]: in function 'require'
        ...share/nvim/rocks/share/lua/5.1/rocks/config/internal.lua:74: in function 'get_rocks_toml'
        ...share/nvim/rocks/share/lua/5.1/rocks/config/internal.lua:113: in function 'get_user_rocks'
        .../luarocks/rocks-5.1/rocks.nvim/2.36.1-1/plugin/rocks.lua:61: in main chunk
```

mrcjkb:
> [!IMPORTANT]
>
> This PR contains 2 commits (one fix and one feature that are related).
> It should be rebased, not squashed.

Fixes https://github.com/nvim-neorocks/rocks-treesitter.nvim/issues/23
Fixes #487
Fixes #342
Closes #441

### DIFF
--- a/installer.lua
+++ b/installer.lua
@@ -369,15 +369,34 @@ local function install()
                 return
             end
 
-            vim.notify("Installing rocks.nvim...")
-            local sc = vim.system({
+            local rocks_binaries_supported_arch_map = {
+                Darwin = {
+                    arm64 = "macosx-aarch64",
+                    aarch64 = "macosx-aarch64",
+                    x86_64 = "macosx-x86_64",
+                },
+                Linux = {
+                    x86_64 = "linux-x86_64",
+                },
+                Windows_NT = {
+                    x86_64 = "win32-x86_64",
+                },
+            }
+            local uname = vim.uv.os_uname()
+            local supported_arch = rocks_binaries_supported_arch_map[uname.sysname][uname.machine]
+
+            local install_cmd = {
                 luarocks_binary,
                 "--lua-version=5.1",
                 "--tree=" .. install_path,
-                "--server='https://nvim-neorocks.github.io/rocks-binaries/'",
                 "install",
                 "rocks.nvim",
-            }):wait()
+            }
+            if supported_arch then
+                table.insert(install_cmd, 4, "--server='https://nvim-neorocks.github.io/rocks-binaries/'")
+            end
+            vim.notify("Installing rocks.nvim...")
+            local sc = vim.system(install_cmd):wait()
 
             if sc.code ~= 0 then
                 notify_output("Installing rocks.nvim failed:", sc, vim.log.levels.ERROR)
--- a/bootstrap.lua
+++ b/bootstrap.lua
@@ -108,16 +108,35 @@ end
 
 assert(set_up_luarocks(temp_luarocks_path), "failed to install luarocks! Please try again :)")
 
-vim.notify("Installing rocks.nvim...")
+local rocks_binaries_supported_arch_map = {
+    Darwin = {
+        arm64 = "macosx-aarch64",
+        aarch64 = "macosx-aarch64",
+        x86_64 = "macosx-x86_64",
+    },
+    Linux = {
+        x86_64 = "linux-x86_64",
+    },
+    Windows_NT = {
+        x86_64 = "win32-x86_64",
+    },
+}
+local uname = vim.uv.os_uname()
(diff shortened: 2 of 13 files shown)


installer.lua
```diff
@@ -369,15 +369,34 @@ local function install()
...
-                "--server='https://nvim-neorocks.github.io/rocks-binaries/'",
                 "install",
                 "rocks.nvim",
-            }):wait()
+            }
+            if supported_arch then
+                table.insert(install_cmd, 4, "--server='https://nvim-neorocks.github.io/rocks-binaries/'")
+            end
+            vim.notify("Installing rocks.nvim...")
+            local sc = vim.system(install_cmd):wait()
```
alexanderjeurissen, line 399: This logic is repeated and also present in `bootstrap.lua`
Might be worthwhile to extract this to a separate Lua module to keep things DRY

mrcjkb: Thanks for the suggestion. The bootstrap and install scripts are standalone and can't make use of module imports. That's why they have a lot of repeated logic.

PR #500, opened 2024-08-06
feat(api): emit user events for cache updates

mrcjkb:
Needed for 

- https://github.com/nvim-neorocks/rocks-edit.nvim/pull/5


mrcjkb:
Actually, perhaps a single `CachePopulated` event with info about the type of cache is better.

PR #509, opened 2024-08-19
fix(windows): use junctions instead of symlinks

mrcjkb:
See #506


mrcjkb:
Auto-merging without review, as this is trivial.

PR #517, opened 2024-08-24
fix(windows): rocks.toml getting wiped on save

### LINKED ISSUE #516: Windows: rocks.toml contents being erased on update/sync/install
I actually found this yesterday while figuring out #513 but I wanted to investigate further before opening an issue.

rocks.toml's contents gets erased whenever it gets written to (e.g. with `:Rocks install`). This means that if you do `:Rocks sync`, rocks uninstalls itself (lol) as it's not in the toml file.

rocks.log is also empty.

image

I'm struggling to narrow down the issue, but when I run `:Rocks install rocks.nvim`, I see the following in process monitor:

image

image

As it opens a pipe, I can assume it's happening in `fs.lua` in the `fs.write_file` function, where it opens a pipe. It's not immediately obvious to me why it might not be writing to the file though. Could be a problem with whatever is calling `fs.write_file`.

I'll keep looking but I am struggling to figure out what's happening.

MeguminSama:
On Windows, `rocks.toml` was getting erased every time rocks tried to write to it.

It seems to have been caused by using pipes (perhaps because the behaviour on Linux is different to that of Windows).

This PR removes the pipe and uses `uv.fs_write(file, ...)` instead.

Resolves #516

### DIFF
--- a/lua/rocks/fs.lua
+++ b/lua/rocks/fs.lua
@@ -48,10 +48,7 @@ function fs.write_file(location, mode, contents, callback)
     -- mode for the group and others
     uv.fs_open(location, mode, tonumber("644", 8), function(err, file)
         if file and not err then
-            local file_pipe = uv.new_pipe(false)
-            ---@cast file_pipe uv_pipe_t
-            uv.pipe_open(file_pipe, file)
-            uv.write(file_pipe, contents, function(write_err)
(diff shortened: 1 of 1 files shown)


NTBBloodbath (approved):

mrcjkb (approved):
thanks 🙏🙏🙏

PR #523, opened 2024-08-31
fix: adapt to nvim 0.11 breaking `vim.version.parse` change

### LINKED ISSUE #522: [nightly/nvim 0.11]: `vim.version.parse` breaking API change
`vim.version.parse` has been changed to return `nil` instead of throwing an error if the argument can't be parsed.
This is a breaking change that affects rocks.nvim.

mrcjkb:
Closes #522

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -80,16 +80,32 @@
         "type": "github"
       }
     },
+    "flake-compat_6": {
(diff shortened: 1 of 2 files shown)


mrcjkb:
Auto-merging as this is trivial.

PR #562, opened 2024-10-17
fix: add tree-sitter parsers to the runtimepath asap

### LINKED ISSUE #561: Parser dependencies don't start automatically
On some parsers (I currently found `astro` and `php` to be affected) highlighting doesn't work when you open a file. It sometimes starts working after restarting treesitter or switching a file, but sometimes doesn't.

Easiest way to reproduce:

```php
<?php

echo "Hello World!";
```

When you open this file with treesitter highlighting enabled, it looks like this:

image

However, after running `:lua vim.treesitter.stop(); vim.treesitter.start()`:

image

mrcjkb:
Fixes #561

### DIFF
--- a/lua/rocks/adapter.lua
+++ b/lua/rocks/adapter.lua
@@ -154,20 +154,28 @@ function adapter.init_site_symlink_sync(rock)
 end
 
 --- Loop over the installed rocks and create symlinks in site/pack/luarocks/opt,
---- so that rtp paths like 'autoload' and 'color' are available before rocks.nvim
+--- so that colorschemes are available before rocks.nvim
 --- has initialised.
 local function init_site_symlinks_async()
     local state = require("rocks.state")
-    for _, rock in pairs(state.installed_rocks()) do
-        init_site_symlink_async(rock)
-        -- Since we're invoking this in the :h load-plugins phase of the startup sequence,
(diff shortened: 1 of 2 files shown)


vhyrro (approved):

lua/rocks/adapter.lua
```diff
@@ -154,20 +154,28 @@ function adapter.init_site_symlink_sync(rock)
...
+        ---@param rock Rock
+        :filter(function(_, rock)
+            return not vim.startswith(rock.name, "tree-sitter-")
+        end)
+        ---@param rock Rock
+        :each(function(_, rock)
+            init_site_symlink_async(rock)
+            -- Make autoload scripts available
+            -- Since we're invoking this in the :h load-plugins phase of the startup sequence,
+            -- this packadd! call won't result in any scripts being sourced.
```
mrcjkb, line 172: Discussion for another issue:
Perhaps we don't actually need this.
It could potentially lead to the subtle issues, because the packadd is delayed (it's async). Neovim only searches for colorschemes in `pack/*/opt`, but not autoload scripts or tree-sitter parsers, which is why I had added this.

But perhaps it's better for now not to support `autoload` scripts as dependencies. It's not really common in the lua ecosystem anyway.

PR #565, opened 2024-10-19
feat(config): `import` option for splitting rocks toml configs

simifalaye:
Addresses: https://github.com/nvim-neorocks/rocks.nvim/issues/240

This PR aims to add support for config importing/inclusion for the purpose of modularizing your rocks.toml and/or supporting local configuration that is outside of version control.

@vhyrro Had mention making this into a separate module but I think configuration imports is a fairly standard/core feature, so I wanted to implement it right in rocks.nvim. It also seemed easier to deal with the complexities of allowing core rocks.nvim and other modules interact with configuration as if it is one table but ensuring that we write all configuration back to the same files that they came from after modification.

Essentially, we'd need to make a breaking change to the handler API to support a list of `MutRocksTomlRef` so that modifications could be isolated to each file. To avoid this, I implemented this using a metatable which contains all the configuration tables associated with each config file and then uses the `__index` and `__newindex` methods to get the correct table for modification. For the config module, the imports are just merged into the base configuration since that isn't being written back to the config files.

*Side Note*: There seems to be an issue in toml-edit where toml lists don't properly convert to a lua table when you use the `parse` function. Only string key tables seem to work for the `parse` function. I may open an issue in the repo unless there's something I'm missing here.


mrcjkb:
This looks like a great addition! :pray: 

I haven't had a chance to look at it in detail yet, but I'll see if I can find some time in the next few days.

lua/rocks/fs.lua
```diff
@@ -36,6 +36,51 @@ function fs.file_exists(location)
...
+function fs.joinpath(...)
+    local sep = package.config:sub(1, 1) -- Get the path separator
+    local args = { ... }
+    return table.concat(args, sep)
+end
+
+--- Get the directory of a file path
+---@param path string
+---@return string
+function fs.dirname(path)
```
mrcjkb, line 67: issue: we just use `vim.fs.dirname`

lua/rocks/fs.lua
```diff
@@ -36,6 +36,51 @@ function fs.file_exists(location)
     return false
 end
 
+--- Expand environment variables and tilde in the path string
+---@param path_str string
+---@return string
+function fs.expand_path(path_str)
```
mrcjkb, line 42: question: isn't this achievable with `vim.fn.expand`?

lua/rocks/operations/helpers.lua
```diff
@@ -29,11 +29,111 @@ local helpers = {}
 
 helpers.semaphore = nio.control.semaphore(1)
 
+---@class MultiMutRocksTomlWrapper
+---@field cache table<string, MultiMutRocksTomlWrapper> Cache for nested metatables
+---@field tables MutRocksTomlRef[] Rocks toml config metatables
+---@field paths? string[] File paths associated the rocks toml configs
```
mrcjkb, line 35:
```suggestion
---@field paths? string[] File paths associated with the rocks toml configs
```

lua/rocks/operations/helpers.lua
```diff
@@ -29,11 +29,111 @@ local helpers = {}
...
+        if tbl[key] ~= nil then
+            tbl[key] = value
+            return
+        end
+    end
+    -- If key not found in any table, add it to the first table
+    self.tables[1][key] = value
+end
+
+--- Write to all rocks toml config files
```
mrcjkb, line 74: suggestion: we should hint that this function needs to be called in an async context for the write to complete before it returns.

```suggestion
---@type async fun()
--- Write to all rocks toml config files
```

lua/rocks/operations/helpers.lua
```diff
@@ -29,11 +29,111 @@ local helpers = {}
 
 helpers.semaphore = nio.control.semaphore(1)
 
+---@class MultiMutRocksTomlWrapper
+---@field cache table<string, MultiMutRocksTomlWrapper> Cache for nested metatables
+---@field tables MutRocksTomlRef[] Rocks toml config metatables
```
mrcjkb, line 34: suggestion: if the paths are associated with the tables, let's make this a `table<string, MutRocksTomlRef>`.

lua/rocks/operations/helpers.lua
```diff
@@ -29,11 +29,111 @@ local helpers = {}
...
+    end
+    for i, path in ipairs(self.paths) do
+        if self.tables[i] then
+            fs.write_file_await(path, "w", tostring(self.tables[i]))
+        end
+    end
+end
+
+--- Function to create a new wrapper
+---@param nested boolean Whether this is a nested metatable
```
mrcjkb, line 87: issue: a boolean parameter is often a sign that this should be split into separate functions. In this case, I'd suggest `new` and `new_nested`.

lua/rocks/operations/helpers.lua
```diff
@@ -29,11 +29,111 @@ local helpers = {}
...
+---@param tables MutRocksTomlRef[] Rocks toml config metatables
+---@param paths? string[] File paths associated the rocks toml configs
```
mrcjkb, line 89: issue: instead of separate params for paths and tables, we can pass in `refs table<string, MutRocksTomlRef> | MutRocksTomlRef[]` and use `vim.tbl_isempty`.

lua/rocks/operations/helpers.lua
```diff
@@ -29,11 +29,111 @@ local helpers = {}
...
+---@return MultiMutRocksTomlWrapper
+function MultiMutRocksTomlWrapper.new(nested, tables, paths)
+    if #tables < 1 then
+        return {}
+    end
+    if not nested and #tables ~= #paths then
```
mrcjkb, line 95: question: is there a use case for having different numbers of paths and tables? It sounds like it would be a bug to me (hence my other recommendations).

lua/rocks/operations/helpers.lua
```diff
@@ -29,11 +29,111 @@ local helpers = {}
 
 helpers.semaphore = nio.control.semaphore(1)
 
+---@class MultiMutRocksTomlWrapper
```
mrcjkb, line 32: suggestion: I think this can go in its own module.

lua/rocks/operations/helpers.lua
```diff
@@ -29,11 +29,111 @@ local helpers = {}
...
+        end
+        visited[file_path] = true
+
+        local config_file = fs.read_or_create(file_path, default)
+        local rocks_toml = require("toml_edit").parse(config_file)
+        table.insert(rocks_tomls, rocks_toml)
+        table.insert(paths, file_path)
+
+        -- Follow import paths
+        -- TODO: Remove call to get_imports, this is needed because toml-edit doesn't
```
mrcjkb, line 123: issue: I'm personally not a fan of todo comments, as they tend to get forgotten (the ones we do have here are quite old).
Let's move this to an issue (I guess in toml-edit).

lua/rocks/operations/helpers.lua
```diff
@@ -29,11 +29,111 @@ local helpers = {}
...
+        return {}
+    end
+    local self = { cache = {}, tables = tables, paths = paths }
+    setmetatable(self, MultiMutRocksTomlWrapper)
+    return self
+end
+
 ---Decode the user rocks from rocks.toml, creating a default config file if it does not exist
----@return MutRocksTomlRef
+---@return MultiMutRocksTomlWrapper
```
mrcjkb, line 104: suggestion: it appears that every call to this function casts the return value to a `MutRocksTomlRef`.
So we should just do the casting in this function.

mrcjkb (changes requested):

lua/rocks/config/internal.lua
```diff
@@ -70,25 +70,59 @@ local default_config = {
...
+                                name = name,
+                                version = data,
+                            }
+                        else
+                            rt[key][name].name = name
+                        end
                     end
                 end
             end
+            -- Merge, giving preference to existing config before imports
```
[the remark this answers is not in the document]
simifalaye: Yeah, you're right. You'd want imports to have preference over the base config so that you could have local config override scm config if you wanted to do that. Since this is a valid use case, we may now want to raise an error. I'll change it to give preference to the imported config.

lua/rocks/fs.lua
simifalaye: Yeah, I was going to do that at first but the issue was that you would need to vim.schedule it to run in the main event loop since it's a vimscript function right? Then, you need to wait to get the result. Could use a future if it's in an async context but this would be needed for the config which is not async. Is there a better way to do it with the vim.fn.expand function?

lua/rocks/fs.lua
```diff
@@ -36,6 +36,51 @@ function fs.file_exists(location)
...
+    if home then
+        path = path:gsub("^~", home)
+    end
+    return path
+end
+
+--- Join paths using the OS specific path separator
+---@vararg string
+---@return string
+function fs.joinpath(...)
```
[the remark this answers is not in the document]
simifalaye: Yup, not sure why I included it

lua/rocks/config/internal.lua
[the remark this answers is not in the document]
mrcjkb: The reason I'm not sure if we want to support overrides is because it increases the complexity: When updating plugins that exist in multiple configs, you have to keep track of which toml file to write to.
There might be other subtle things I'm not considering here. 

I'm not sure if the ability to override scm configs is strong enough of a use case, which is why I'd say YAGNI for now.

lua/rocks/operations/helpers.lua
simifalaye: Did this at first but wanted to be able to control what order the tables are parsed in so that preference could be given to the base configuration (though, now I will change the order so that imports are given preference). From my understanding, the order isn't guaranteed when you use a string-keyed table. Any ideas on how I could give preference to a specific config file when matching the keys?

lua/rocks/operations/helpers.lua
simifalaye: Nope, it should never happen but it's just on consequences of using a list table instead of a string-keyed table. See my other comment: Just wondering how best to address defining a config preference order in when using the path as the table key.

lua/rocks/config/internal.lua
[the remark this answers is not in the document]
simifalaye: I think (correct me if I'm wrong) because both the config and the operations respect config preferences, you will always write back to the config that has higher preference if it exists in both configs. The config.internal module merges all the configurations in order and the operations.helpers module orders the config tables by preference. So, I think it shouldn't add complexity but I may be wrong so I'll leave that up to you to decide.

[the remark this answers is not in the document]
mrcjkb: That's probably correct. But I'm still not sure if it has any implications I'm not foreseeing, which is why I'd rather not support overrides until people ask for it.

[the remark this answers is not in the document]
mrcjkb: come to think about it, we should probably add some tests for this feature.

lua/rocks/fs.lua
simifalaye: Hmm, I'll just make a `get_absolute_path_await` which wraps the original function with a future `set` and `wait` and call that one from the async context

lua/rocks/operations/helpers.lua
mrcjkb: Does the order still matter if we treat conflicts as an error? 
If so, we can make it a list of tuples:

```lua
---@class MutRocksTomlRefWithPath
---@field path string
---@field ref MutRocksTomlRef

-- ...
---@field tables MutRocksTomlRefWithPath[]
```

simifalaye: Order won't matter if we don't support overrides and just treat conflicts as errors. Also, yeah sounds good on the tuple, just thought of this too today and I'm changing the implementation as we speak. I feel like it might be good to just implement with overrides in mind in the future because I think that will be the next logically request when config importing is supported.

simifalaye:
Btw @mrcjkb I opened this issue for toml_edit.lua (the lua library wrapper and not the original toml-edit). I'll take a look at it later but my rust is rusty so it will take me some time. Might be a quick fix for someone who is very familiar since it looks like you'd just need to match on a new toml-edit value type in the parse function (`toml_edit::Value::Array`)

lua/rocks/fs.lua
simifalaye: Decided to extract out a function for parsing the rocks_toml which is usable by both the config and helpers modules, hence, I kept the implementation of `expand_path` which doesn't depend on the main loop. Let me know if you'd still prefer that I use `vim.fn.expand`

lua/rocks/operations/helpers.lua
simifalaye: Finally had some time to work on the toml-edit change last night: https://github.com/nvim-neorocks/toml-edit.lua/pull/38

mrcjkb: Merged and released. Shall we update the dependency in this PR?

mrcjkb (changes requested):
CI is failing on stylua.

lua/rocks/operations/helpers.lua
simifalaye: Done

simifalaye:
@mrcjkb Updated the flake.lock also but not sure if that's needed now for the CI? Feel free to take over the PR to get it through as I'm just new to nix. I believe all of the code, unit tests and style stuff should be addressed now but the CI might need some other things to get new toml-edit changes.

mrcjkb (dismissed):

mrcjkb:
Merging into the `split-rocks-toml` branch, so I can fix the build there.

mrcjkb:
@simifalaye I just noticed that since this PR, our integration test on Ubuntu is failing with

```
E5113: Error while calling lua chunk: ....nvim/rocks.nvim/rocks/share/lua/5.1/luarocks/loader.lua:104: error loading module 'toml_edit' from file '/home/runner/work/rocks.nvim/rocks.nvim/rocks/lib/lua/5.1/toml_edit.so':
	/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /home/runner/work/rocks.nvim/rocks.nvim/rocks/lib/lua/5.1/toml_edit.so)
```

any idea what could be going on?

mrcjkb:
nevermind, I've upgraded the Ubuntu runner and it's fixed.

PR #577, opened 2024-10-31
feat(api): Add support for install options

simifalaye:
As per this discussion: https://github.com/nvim-neorocks/rocks.nvim/discussions/575

This PR adds support for install options to the api, specifically, the option to specify a relative config path to install a new rock to. The config path is then added to the imports of the base configuration file.


lua/rocks/operations/add.lua
```diff
@@ -87,16 +88,16 @@ add.add = function(arg_list, callback, opts)
...
+            local user_rocks = helpers.parse_rocks_toml(opts.config_path)
+            local handler = handlers.get_install_handler_callback(user_rocks --[[@as MutRocksTomlRef]], arg_list)
             if type(handler) == "function" then
                 local function report_progress(message)
                     progress_handle:report({
                         message = message,
                     })
                 end
                 handler(report_progress, report_error, helpers.manage_rock_stub)
-                user_rocks:write()
```
mrcjkb, line 99: question: is this change necessary?
I find `user_rocks:write()` more readable than `user_rocks()`.

lua/rocks/operations/helpers/init.lua
```diff
@@ -30,21 +30,61 @@ local helpers = {}
...
+            require("toml_edit").parse(fs.read_or_create(config.config_path, constants.DEFAULT_CONFIG))
+        -- Check to see if the path provided was the base config path, if so, just return that
+        if absolute_config_path == config.config_path then
+            return multi_mut_rocks_toml_wrapper.new({ { config = base_rocks_toml, path = config.config_path } })
+        end
+
+        -- For non-base configs, add it to the list of imports in the base config
+        if base_rocks_toml.import then
+            local i = 0
+            while true do
```
mrcjkb, line 51: suggestion: instead of `while true` with a `break`, use `do ... while`.

lua/rocks/operations/pin.lua
```diff
@@ -24,7 +24,7 @@ pin.pin = function(rock_name)
     nio.run(function()
         helpers.semaphore.with(function()
             local user_config = helpers.parse_rocks_toml()
-            local rocks_key, user_rock = helpers.get_rock_and_key(user_config, rock_name)
+            local rocks_key, user_rock = helpers.get_rock_and_key(user_config --[[@as MutRocksTomlRef]], rock_name)
```
mrcjkb, line 27: suggestion: I believe you can use an [`---@overload` annotation](https://luals.github.io/wiki/annotations/#overload) to avoid having to cast.

another (preferable) option (that I haven't looked into due to lack of time) could be to extract a shared interface.

mrcjkb (changes requested):

lua/rocks/operations/add.lua
simifalaye: It was for the purpose of removing the use of non-meta fields/methods from the `MultiMutRocksTomlWrapper` type so that it would be logically equivalent to the `MutRocksTomlRef` type. Right now, the two types are almost equivalent in that if you access an index from the wrapper object, you'll only get whatever is defined in the table object types. But there is an additional "write" field in the wrapper type which will be given preference over the same key in the tables (see the `__index` I added in the last PR). This could create issues in the future as there could be clashing field names between the config and the wrapper class. Since this is just meant to be helper class that should represent the config type, I didn't want there to be any consideration required or possible issues in the future when adding/renaming fields in the config file just for a helper class.

mrcjkb: I see. To me, it feels very unlikely that the table object type would have a field called `write`. But to be on the safer side, we could call the field something like `_write_await` (something highly unlikely to be used, and also indicating that it's internal with the `_` prefix and hinting the async context with the `_await` suffix).

We could then document in the api module that rocks.toml keys should not have  `_` prefixes, thus reserving them for potential meta  fields in the future.
Theoretically, this would be a breaking change, but I wouldn't treat it as one in this case, because I know for a fact that no rocks.nvim extension adds `_`-prefixed keys to the rocks.toml schema.

@vhyrro any thoughts on this?

simifalaye: Yeah, I was also thinking about using the `_` prefix to denote that it's an internal method but I still wasn't sure if that would still pose a risk. I'm fine with whatever you guys decide!

Another thought could be that the `__call` metamethod could accept a function of type `fun(configs: MutRocksTomlRefWithPath[])` and pass in the internal config table to allow then method you pass in to act on them as individual config tables. Then you could have an external function (as part of the `multi_mut_rocks_toml_wrapper` module) called something like `write_rocks_toml` which would take in a wrapper object and pass a function to it which writes each config object to their respective paths?

lua/rocks/api/init.lua
```diff
@@ -188,12 +188,15 @@ function api.source_runtime_dir(dir)
     require("rtp_nvim").source_rtp_dir(dir)
 end
 
+---@class InstallRockOpts: rocks.AddOpts
```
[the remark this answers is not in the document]
simifalaye: Done

lua/rocks/operations/helpers/multi_mut_rocks_toml_wrapper.lua
```diff
@@ -53,13 +58,10 @@ MultiMutRocksTomlWrapper.__newindex = function(self, key, value)
 end
 
 --- Write to all rocks toml config files in an async context
```
simifalaye, line 60: Need to change this documentation

lua/rocks/operations/pin.lua
```diff
@@ -38,7 +38,7 @@ pin.pin = function(rock_name)
             end
             user_config[rocks_key][rock_name].pin = true
             local version = user_config[rocks_key][rock_name].version
-            user_config:write()
+            user_config()
```
simifalaye, line 41: Forgot to change this to the helper function

lua/rocks/operations/add.lua
simifalaye: Done

simifalaye:
Note @mrcjkb, if i'm seeing this correctly, toml-edit is being pulled in as a nix package and you are the maintainer? If so, toml-edit-0.6.0 will need to be added since only 0.5.0 exists

mrcjkb:
> Note @mrcjkb, if i'm seeing this correctly, toml-edit is being pulled in as a nix package and you are the maintainer? If so, toml-edit-0.6.0 will need to be added since only 0.5.0 exists

https://github.com/NixOS/nixpkgs/pull/354096#issuecomment-2461524002

mrcjkb:
Needs a rebase and conflict resolution (I've bumped the `toml-edit` dependency to `>= 0.6.1`).

simifalaye:
Should be done. Please review again as I also corrected some things (changing the internal code to use the new api options also)

PR #590, opened 2024-11-18
fix: make sure config_path option for installs updates correct config

simifalaye:
Fixes an issue with the config_path option on installs (NOTE: this only affected the new config_path feature added last PR and not anything else). This was meant to be committed as part of the last PR but I had stashed these changes temporarily while testing a slightly different implementation and then forgot to re-apply the stashed changes afterwards (I apologize for this).

The original intention was that when you provided a config_path to the `parse_rocks_toml` function, it would add the import to the base config and then provide you a wrapper object which wrapped both the base config and the newly created import config so that you could write both at once when done updating. The issue with this is that the goal of the config_path option is so that you could make changes ONLY to the specified config file but since I was providing a wrapped object, it meant that despite the imported config having preference, it could end up preferring the base config for writing if the nested table paths didn't exist in the import config (hence defeating the purpose of this feature). I realized this and changed it locally to asynchronously write the base config when the imports have been updated and only return the rocks_toml for the specified config_path so that all modifications would be guaranteed to be only done on the config_path but this was not committed unfortunately.

### DIFF
--- a/lua/rocks/operations/helpers/init.lua
+++ b/lua/rocks/operations/helpers/init.lua
@@ -45,7 +45,7 @@ function helpers.parse_rocks_toml(config_path)
             return multi_mut_rocks_toml_wrapper.new({ { config = base_rocks_toml, path = config.config_path } })
         end
 
-        -- For non-base configs, add it to the list of imports in the base config
+        -- For non-base configs, add it to the list of imports in the base config and write async
         if base_rocks_toml.import then
             local i = 0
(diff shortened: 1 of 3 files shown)


mrcjkb (changes requested):
Thanks :pray: 
CI (luacheck) is failing on some unused variables in the mut_rocks_toml_reapper module.

mrcjkb (approved):

PR #613, opened 2024-12-20
feat: option to auto-sync on startup

### LINKED ISSUE #293: [Feature] Prompt to sync rocks if plugins cannot be loaded at startup
With https://github.com/nvim-neorocks/rocks.nvim/pull/292, rocks.nvim shows a warning if plugins could not be loaded at startup.

We could also prompt to run `:Rocks sync`.

mrcjkb:
Closes #293


NTBBloodbath (approved):
LGTM

PR #617, opened 2024-12-26
docs(readme): make bootstrap snippet work well with lua-language-server

saccarosium:
Hi,
this PR tries to:
- make the bootstrapping snippet work a bit nicer with lua-language-server. lua lsp will complain about the use of `vim.fn.stdpath` in `vim.fs.joinpath` since it can return `string[]` and will throw a warning. The fix is telling it to treat the `stdpath` call as a string.
- make the bootstrapping snippet less intimidating by reducing is sice in LOC
- remove redundant `vim.fs.normalize` call (`vim.fs.joinpath` already does that`)
- move the `vim.v.shell_error` check in the same branch as a "sanity check". It is nitpicking but, as is it right now, this can be trigger by _any_ `vim.fn.system` or `:!` gone wrong in the user configuration and moving it to the same branch as the `vim.fn.system` call that clones the repo this can make life much saner to user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ one may use the bootstrapping script. Place the following script into your `init
 ```lua
 do
     -- Specifies where to install/use rocks.nvim
-    local install_location = vim.fs.joinpath(vim.fn.stdpath("data"), "rocks")
+    local install_location = vim.fs.joinpath(vim.fn.stdpath("data") --[[@as string]], "rocks")
 
     -- Set up configuration options related to rocks.nvim (recommended to leave as default)
     local rocks_config = {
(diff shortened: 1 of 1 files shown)


mrcjkb (changes requested):
Hey 👋 
Thanks for your PR 🙏

> remove redundant vim.fs.normalize call (vim.fs.joinpath already does that`)

That is incorrect. `joinpath` (as of writing) only does `gsub('//+', '/')`, which is less than `normalize`.

Please also revert your change to the `joinpath` arguments. Right now, `joinpath` adds a `/` on all platforms, but that could change.

Please also refer to the review checklist, specifically regarding the use of conventional commits.

saccarosium:
> That is incorrect. joinpath (as of writing) only does gsub('//+', '/'), which is less than normalize.

Interesting, I was basing this on the documentation. I didn't actually read the implementation. I should have done it...

> Concatenate directories and/or file paths into a single path with
> normalization (e.g., `"foo/"` and `"bar"` get joined to `"foo/bar"`)
> -- `:h vim.fs.joinpath`

mrcjkb (approved):

PR #624, opened 2025-01-31
docs(readme): add warning about rewrite

mrcjkb:


mrcjkb:
integration test failure is unrelated.

PR #630, opened 2025-02-11
refactor: removed docgen by inlining it into the Makefile

teto:
this feels less magical, I was wondering what was this 'docgen' program. My starting problem was that I wanted to run the doc generator from nixpkgs as well, which was harder before this PR

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -111,7 +111,8 @@
             docgen = {
               enable = true;
               name = "docgen";
-              entry = "${pkgs.docgen}/bin/docgen";
+              entry = "make docgen";
+              extraPackages = [inputs.vimcats.packages.${system}.default];
               files = "\\.(lua)$";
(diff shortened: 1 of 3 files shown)


mrcjkb (changes requested):
todo: I'm fine with this change, but we'll need to update the git-hooks check.

flake.nix
```diff
@@ -111,7 +111,7 @@
             docgen = {
               enable = true;
               name = "docgen";
-              entry = "${pkgs.docgen}/bin/docgen";
+              entry = "make docgen";
```
teto, line 114: is this the git hook check or do you mean sthg else ?

mrcjkb: Yep. The hook appears to be working but or some reason it's deleting all the docs :thinking:

mrcjkb (approved):

PR #712, opened 2026-04-13
fix: adapt to breaking luarocks change in package search

### LINKED ISSUE #711: [Bug] `:Rocks search` only completes dev/scm packages
Something in the `luarocks search` CLI has changed (probably since the Teal rewrite?), so that when you pass in the `--dev` flag, it only searches dev packages and excludes the main manifest.
As a result, rocks.nvim no longer includes non-dev packages in its search results.

mrcjkb:
Fixes #711

### DIFF
--- a/nix/plugin-overlay.nix
+++ b/nix/plugin-overlay.nix
@@ -83,7 +83,7 @@ in {
       viAlias = false;
       vimAlias = false;
       plugins = [
-        final.vimPlugins.rocks-nvim
+        (final.vimPlugins.rocks-nvim.overrideAttrs {doCheck = false;})
       ];
     };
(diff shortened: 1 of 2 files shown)


nix/plugin-overlay.nix
```diff
@@ -83,7 +83,7 @@ in {
       viAlias = false;
       vimAlias = false;
       plugins = [
-        final.vimPlugins.rocks-nvim
+        (final.vimPlugins.rocks-nvim.overrideAttrs {doCheck = false;})
```
mrcjkb, line 86: no need to run `vimPlugin` checks here.

kaistian:
Works great, thank you.

PR #716, opened 2026-05-01
feat: change log folder so one can use :log from nightly

teto:
neovim now has `:log `. This changes the log folder so one can do `:log rocks` . Not sure of the implications: `:Rocks log` keeps working.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -50,7 +50,7 @@ jobs:
           nvim -u .github/resources/init-integration.lua -l .github/resources/install-colorscheme.lua +q
           ls rocks
           # Will fail if rocks.log does not exist
-          nvim --headless -u NONE "+lua print(table.concat(vim.fn.readfile(vim.fs.joinpath(vim.fn.stdpath('state'), 'rocks.log')), '\n'))" +q
+          nvim --headless -u NONE "+lua print(table.concat(vim.fn.readfile(vim.fs.joinpath(vim.fn.stdpath('log'), 'rocks.log')), '\n'))" +q
           echo "vim.cmd.colorscheme('sweetie')" >> .github/resources/init-integration.lua
           echo "vim.cmd.e('success')" >> .github/resources/init-integration.lua
           nvim -u .github/resources/init-integration.lua -c 'set noswapfile' +wq
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -50,7 +50,7 @@ jobs:
           nvim -u .github/resources/init-integration.lua -l .github/resources/install-colorscheme.lua +q
           ls rocks
           # Will fail if rocks.log does not exist
-          cat rocks/rocks.log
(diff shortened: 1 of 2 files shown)


lua/rocks/log.lua
```diff
@@ -44,7 +44,7 @@ end
 ---@package
 ---@return string filepath
 function log.get_logfile()
-    return vim.fs.joinpath(require("rocks.config.internal").rocks_path, "rocks.log")
+    return vim.fs.joinpath(vim.fn.stdpath("log"), "rocks.log")
```
mrcjkb, line 47: question: when was this introduced? Perhaps we should fall back to the `rocks_path` if `stdpath` returns `nil` (for backward compatibility with older Nvim versions).

mrcjkb (changes requested):

teto:
```
commit 843a7223e77d63be496e6bcfbe1fca89b7010266
Merge: 94c0e54e5b 8715877417
Author: Justin M. Keyes <justinkz@gmail.com>
Date:   Fri Apr 10 12:38:38 2026 -0400

    Merge #38501 :log, stdpath("state")/logs

commit 8715877417b5c5fee98e6cac9e4e60ca08050048
Author: Olivia Kinnear <git@superatomic.dev>
Date:   Sat Mar 28 22:07:04 2026 -0500

    feat(ex): add `:log` command
```

stdpath('log') is available since 2023 at least, it just pointed at another folder. Since rocks resets the log on every launch, it should be fine to just change where it lives.

mrcjkb (changes requested):
The integration test in CI needs to be updated.
It's failing with:

```
cat: rocks/rocks.log: No such file or directory
```

mrcjkb:
> The integration test in CI needs to be updated. It's failing with:
> 
> ```
> cat: rocks/rocks.log: No such file or directory
> ```

Same for Windows: 

```
cat: /c/Users/runneradmin/.local/state/nvim/rocks.log: No such file or directory
```

.github/workflows/integration.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           nvim -u .github/resources/init-integration.lua -l .github/resources/install-colorscheme.lua +q
           ls rocks
           # Will fail if rocks.log does not exist
-          cat rocks/rocks.log
+          nvim --headless -u NONE "+lua print(table.concat(vim.fn.readfile(vim.fs.joinpath(vim.fn.stdpath('state'), 'rocks.log')), '\n'))" +q
```
mrcjkb:
```suggestion
          nvim --headless -u NONE "+lua print(table.concat(vim.fn.readfile(vim.fs.joinpath(vim.fn.stdpath('log'), 'rocks.log')), '\n'))" +q
```

mrcjkb (changes requested):
ls: cannot access '/home/runner/.local/state/nvim/logs/': No such file or director

mrcjkb (approved):

teto:
ty

